PR #3, opened 2016-08-05
Activity Log - Users want to see when an action is triggered on the activity log

arnaudbesnier:
- [x] Remove one hook and test again

### DIFF
--- a/routes/actions.js
+++ b/routes/actions.js
@@ -1,0 +1,27 @@
+'use strict';
+var _ = require('lodash');
+var auth = require('../services/auth');
+var ActivityLogLogger = require('../services/activity-log-logger');
+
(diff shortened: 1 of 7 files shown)


routes/actions.js
```diff
@@ -0,0 +1,32 @@
...
+var _ = require('lodash');
+var auth = require('../services/auth');
+var ActivityLogLogger = require('../services/activity-log-logger');
+
+module.exports = function (app, opts) {
+
+  function logActivities(request, response, next) {
+    function logCustomActions() {
+      response.removeListener('finish', logCustomActions);
+      response.removeListener('close', logCustomActions);
```
SeyZ: Can you explain me?

PR #15, opened 2016-09-13
Added : modelsDir is now optional

kevin-frugier:
Since the user can already have registered their models prior to intializing forest, modelsDir must be optional.
In previous implementation, we couldn't provide undefined or null and any dummy value would crash.

Having to provide a valid path (for instance the "root" of the current NodeJS project) was a bad workaround since it would require even non-NodeJS files and would prevent the user from controlling the "require" order.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,15 +21,21 @@ var request = require('superagent');
 var logger = require('./services/logger');
 
 function requireAllModels(Implementation, modelsDir) {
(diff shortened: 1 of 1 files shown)


SeyZ:
Thank you for the PR @kevin-frugier ! We're reviewing it :)

SeyZ:
We merged your pull request after very little changes 👍

PR #16, opened 2016-09-13
Use expressMountParent optional setting to mount forest as a subapp

kevin-frugier:
By default, the express() app used by forest is used as a middleware mounted directly at the root of the parent app :
```
app.use(forestLiana.init({
  modelsDir: __dirname + '/models',  // Your models directory.
  secretKey: 'YOUR-SUPER-SECRET-SECRET-KEY',
  authKey: 'YOUR-SUPER-SECRET-AUTH-KEY', // Choose a secret authentication key.
  ....
}));
```

Forest uses JWT for authentication and relies on `req.user` for the middleware `ensureAuthenticated`

In my case I have other external conditions (like passport) already playing with `req.user`.

In order to keep my REST api (used by my app's frontend) separated from FOREST api, I'd rather use either Express routers or Express sub-apps.

I have quickly refactored the code to show how it could be done.
Instead of mounting all the routes under `/forest/...`, when the option parameter `expressMountParent` is provided, the routes are mounted under `/...` and it's the forest sub-app itself which is mounted under `/forest` in its parent.
This removes for instance the need for the hacky `jwt(...).unless({ path: /^(?!\/forest).*/ })` we see in the code.

In my case my REST api is already in a sub-router under `/api/v1` and I have now moved all `passport` authentication there.
Using forest in it's own sub-app allows for proper separation of duties and avoids middleware clashes.

Here's how I use it : 
```
    var forestLiana = require('forest-express-mongoose');

    var forestApp = forestLiana.init({
        expressMountParent: app,
        secretKey: process.env.FOREST_SECRET,
        authKey: process.env.FOREST_AUTH,
        mongoose: myMongooseRef
    });
```

Feel free to change the name of the setting parameter or other stuff.

I personally wouldn't want to go live in PRODUCTION without such feature.

As a side information, I have tested it with both my frontend and forestadmin without problem and for instance I don't get the DEBUG message anymore from body parser saying `body-parser - body already parsed` since now Forest and my REST api are properly each setting their own.

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,10 +197,20 @@ exports.init = function (Implementation) {
   app.use(bodyParser.json());
 
   // Authentication
-  app.use(jwt({
-    secret: opts.authKey,
-    credentialsRequired: false
-  }).unless({ path: /^(?!\/forest).*/ }));
+  if (!opts.expressMountParent) {
+    //Normal case, app is mounted at root level inside parent expressApp : we use JWT only in /forest paths
+    app.use(jwt({
+        secret: opts.authKey,
+        credentialsRequired: false
+      }).unless({ path: /^(?!\/forest).*/ }));
+  } else {
+    //Subapp case, we'll mount app at /forest level inside parent so we protect every paths
+    app.use(jwt({
+        secret: opts.authKey,
+        credentialsRequired: false
+      })); 
+  }
+  
 
   new SessionRoute(app, opts).perform();
 
@@ -289,6 +299,10 @@ exports.init = function (Implementation) {
       }
     });
 
+  if (opts.expressMountParent) {
+    var parentApp = opts.expressMountParent.parent;
+    opts.expressMountParent.use('/forest', app);
+  }
   return app;
 };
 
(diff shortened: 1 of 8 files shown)


arnaudbesnier:
Thanks for this pull request, we'll review your code this week for sure!
🌳🌳🌳

arnaudbesnier:
We merged your pull request after some changes. You can check the whole change in the following commit: https://github.com/ForestAdmin/forest-express/commit/b02bfcdcfc543a36d3d6b13ca3dfebab9509bb0a

Thanks again!
🌳 🌳 🌳

kevin-frugier:
Thank you very much for merging it. I like the way you factorized the path generation. :+1: 
There is a small double slash issue though : https://github.com/ForestAdmin/forest-express/commit/b02bfcdcfc543a36d3d6b13ca3dfebab9509bb0a#commitcomment-19074521
Nothing too hard to fix :smile: 

Thanks !

arnaudbesnier:
Fix in the next commit: https://github.com/ForestAdmin/forest-express/commit/4ef0d5cf1a80bade09f18561f4fb3996a4daf861

PR #18, opened 2016-09-17
Isolated integrations

SeyZ:


integrations/stripe/setup.js
```diff
@@ -33,7 +33,7 @@ exports.createCollections = function (Implementation, apimap,
     ],
     actions: [{
       name: 'Refund',
-      endpoint: '/forest/stripe_payments/refunds'
+      endpoint: '/forest/' + collectionName + '_stripe_payments/refunds'
```
arnaudbesnier: Thank you for the fix!

PR #20, opened 2016-09-28
Close.io integration

SeyZ:


integrations/close.io/setup.js
```diff
@@ -0,0 +1,67 @@
...
+
+  schema.fields.push({
+    field: 'closeio_lead',
+    type: 'String',
+    reference: collectionName + '_closeio_leads.id',
+    column: null,
+    isSearchable: false,
+    integration: 'close.io'
+  });
+
```
arnaudbesnier: Not integrated in the same way the `refund` action is integrated with Stripe.
Don't know if there is a specific reason.

Not sure that `createField` is the right spot anyway.

routes/closeio.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
```
arnaudbesnier: I think this file is not used anymore; moved in the `integration` directory?

services/closeio-customer-lead-getter.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
```
arnaudbesnier: To Remove?

services/closeio-lead-emails-getter.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
arnaudbesnier: To Remove?

services/closeio-lead-getter.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
```
arnaudbesnier: To Remove?

services/closeio-leads-getter.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
```
arnaudbesnier: To Remove?

integrations/close.io/setup.js
SeyZ: • Agreed with `createField` name.
• Yes, it's not the same as `refund` action on Stripe. The `Create Close.io lead` is added on an existing collection (`mapping`). The `refund` action is added on a smart collection (`stripe.payments`)

PR #27, opened 2016-10-19
Custom Action - Fix the bad endpoints if some actions have the same name

arnaudbesnier:


services/logger.js
```diff
@@ -24,7 +24,7 @@ var CONFIG = {
 
 var TITLE = '[forest] 🌳🌳🌳  ';
 
-var logger = module.exports = new (winston.Logger)({
+module.exports = new (winston.Logger)({
```
arnaudbesnier: Warning removed (unused variable)

PR #50, opened 2017-02-09
Configuration - Catch the error if the modelsDir configured does not exist

arnaudbesnier:


arnaudbesnier:
Fixes https://github.com/ForestAdmin/forest-express-mongoose/issues/40

PR #61, opened 2017-03-24
Make integrations pluggable

### LINKED ISSUE #60: Make the integration system pluggable
Hi,

I'd like to integrate InvoiceNinja into Forest, by taking inspiration on the Stripe integration. After digging around the integration code, I realize it will be difficult for me to do that without forking forest-express.
It seems it could be easy to change https://github.com/ForestAdmin/forest-express/blob/devel/integrations/index.js#L5 to initialize `modules` with values passed through the liana options.

Before I open a pull request, I'd like to know if you would be willing to make the integration system pluggable.

Thanks :-)

louisremi:
fix #60

### DIFF
--- a/integrations/index.js
+++ b/integrations/index.js
@@ -2,7 +2,7 @@
 var fs = require('fs');
 
 function IntegrationChecker(opts) {
-  var modules = [];
+  var modules = opts.integrationModules || [];
 
   fs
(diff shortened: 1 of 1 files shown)


louisremi:
I'm closing this PR, as we're not likely to create a custom integration.

PR #65, opened 2017-04-03
Improve modelsDir error reporting

louisremi:
+ remove useless catch (it's never reached as requireAllModels always resolves)

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,8 +32,8 @@ function requireAllModels(Implementation, modelsDir) {
       })
       .catch(function (error) {
         if (error.code === 'ENOENT') {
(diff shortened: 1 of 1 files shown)


arnaudbesnier:
Hey @louisremi , thanks for the contribution! We have a current PR that fixes the issue available here: https://github.com/ForestAdmin/forest-express/pull/64

arnaudbesnier:
You're right about the catch!
But we also have to avoid an error report if the `/forest` directory does not exist (as I did in the PR).

louisremi:
Perfect!

PR #68, opened 2017-04-04
[-] Performance - Speed up the session response using bcrypt.compare instead of bcrypt.compareSync

louisremi:
_Continuing my efforts to get Forest to run on Amazon Lambda + API Gateway using Claudia.js_

Amazon Lambda is configured by default to timeout after 3s of execution. Using the basic instance, I could never log into Forest because the limit of 3s was systematically reached. I was able to solve my problem by upgrading my instance (3x CPU and RAM), but I wasn't satisfied with the solution, as only the `/session` calls were causing timeouts, and the rest of the API calls were running fast enough.

After a day of investigation, I figured that `bcrypt.compareSync` was the culprit, and that switching to the async `bcrypt.compare` equivalent resulted in 2x performance improvements (!!), using the same instance.

Here's a screenshot of the logs I get when logging-in 4 times with the `bcrypt.compareSync` version:
bcrypt-comparesync

And the same logs using the async `bcrypt.compare` version:
bcrypt-compare

This PR switches to the async `bcrypt.compare` (and also reorganizes the code a bit to throw as early as possible when `opts.authSecret` isn't configured).

NB: using the native module `bcrypt` isn't an option, as it makes it much harder to deploy on Lambda (the module must be cross-compiled for the target environment).

### DIFF
--- a/routes/sessions.js
+++ b/routes/sessions.js
@@ -12,56 +12,61 @@ module.exports = function (app, opts) {
     new AllowedUsersFinder(request.body.renderingId, opts)
       .perform()
       .then(function (allowedUsers) {
+        if (!opts.authSecret) {
+          throw new Error('Your Forest authSecret seems to be missing. Can ' +
+            'you check that you properly set a Forest authSecret in the ' +
+            'Forest initializer?');
+        }
+
         if (allowedUsers.length === 0) {
-          return response.status(401).send({
-            errors: [{
-              detail: 'Forest cannot retrieve any users for the project ' +
-                'you\'re trying to unlock.'
-            }]
-          });
+          throw new Error('Forest cannot retrieve any users for the project ' +
+            'you\'re trying to unlock.');
         }
 
         var user = _.find(allowedUsers, function (allowedUser) {
-          return allowedUser.email === request.body.email &&
-            bcrypt.compareSync(request.body.password, allowedUser.password);
+          return allowedUser.email === request.body.email;
         });
 
-        if (user) {
-          if (opts.authSecret) {
-            var token = jwt.sign({
-              id: user.id,
-              type: 'users',
-              data: {
-                email: user.email,
-                'first_name': user['first_name'],
-                'last_name': user['last_name'],
-                teams: user.teams
-              },
-              relationships: {
-                renderings: {
-                  data: [{
-                    type: 'renderings',
-                    id: request.body.renderingId
-                  }]
-                }
-              }
-            }, opts.authSecret, {
-              expiresIn: '14 days'
-            });
-
-            response.send({ token: token });
-          } else {
-            return response.status(401).send({
-              errors: [{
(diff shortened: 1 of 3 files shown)


routes/sessions.js
```diff
@@ -12,56 +12,61 @@ module.exports = function (app, opts) {
...
-                'you\'re trying to unlock.'
-            }]
-          });
+          throw new Error('Forest cannot retrieve any users for the project ' +
+            'you\'re trying to unlock.');
         }
 
         var user = _.find(allowedUsers, function (allowedUser) {
-          return allowedUser.email === request.body.email &&
-            bcrypt.compareSync(request.body.password, allowedUser.password);
```
Amoki: I think the real optimization is here.
Async bcrypt uses another thread to compute the comparison and doesn't lock the event loop, which is nice but doesn't speed up the process for 1 user

louisremi: I'm not sure what you mean exactly, but bcrypt.compare will never run for more than 1 user anyway.

louisremi: (`bcrypt.compare` only runs if `allowedUser.email === request.body.email`, which can only be true for a single user, as emails are unique)

Amoki: You're right...
So compareSync is surprisingly bad...

SeyZ:
Hey guys, thank you for the PR ! We will take a look deeper! You rock!

routes/sessions.js
```diff
@@ -12,56 +12,61 @@ module.exports = function (app, opts) {
     new AllowedUsersFinder(request.body.renderingId, opts)
       .perform()
       .then(function (allowedUsers) {
+        if (!opts.authSecret) {
```
arnaudbesnier: 👍

louisremi:
Thank you for the review! I'll update the changelog.

Throwing errors is an integral part of flow control when working with promises. You can't just `return res.send(...)` from within a `then` callback if you want your program to stop there: you have to throw if you want to avoid the subsequent `then` callbacks to be executed. 
There are only two alternatives that you don't really want to use:
- The first one is to check that the arguments passed to all your `then` callbacks are always what you expect them to be, but it's quite ugly and doesn't scale well.
- The second is to replace `throw new Error()` with `return new Promise((r, reject) => { reject(); });`, which is only a verbose equivalent of the former code.

Since those expected errors are `catch`ed, they shouldn't appear in error tracking tools. If they do, that might mean the tool is misconfigured.

If you see a more elegant solution that avoids using those errors, please let me know :-)

arnaudbesnier:
Hi @louisremi, I made a copy PR to play with your code and made a few changes: https://github.com/ForestAdmin/forest-express/pull/69

I just changed the change in the changelog and the 401 responses body if there is no error message to keep the same behaviour as before.

arnaudbesnier:
Your contribution is now in now merged, I close this PR.

PR #69, opened 2017-04-10
[-] Performances - Make the password comparison asynchronous on session creation

arnaudbesnier:
This pull request is the same as https://github.com/ForestAdmin/forest-express/pull/68 with some light improvements.


arnaudbesnier:
Thanks to @louisremi for this contribution!

PR #73, opened 2017-04-21
[+] Smart Fields - Developers can now define Smart Fields setters

arnaudbesnier:


arnaudbesnier:
To test promises:

```
'use strict';
const _ = require('lodash');
const models = require('../models');
const Liana = require('forest-express-sequelize');

Liana.collection('user', {
  fields: [{
    field: 'fullName',
    type: 'String',
    get: function (user) {
      if (user.firstName && user.lastName) {
        return `${user.firstName} ${user.lastName}`;
      } else {
        return null;
      }
    },
    set: function (user, value) {
      var names = value.split(' ');
      user.firstName = names[0];
      user.lastName = names[1];

      console.log('== user: ', user);
      return user;
    },
    search: function (query, search) {
      var s = models.sequelize;
      var split = search.split(' ');

      var searchCondition = s.and(
        { firstName: { $ilike: split[0] }},
        { lastName: { $ilike: split[1] }}
      );

      let searchConditions = _.find(query.where.$and, '$or');
      searchConditions.$or.push(searchCondition);
    }
  }, {
    field: 'blu',
    type: 'String',
    get: function (user) {
      return user.email;
    },
    set: function (user, value) {
      // user.email = value;
      return models.user.findAll({ where: { id: 2 }})
        .then(() => {
          user.email = value;
          return user;
        });
    },
    search: function (query, search) {
      var s = models.sequelize;
      var split = search.split(' ');

      var searchCondition = s.and(
        { firstName: { $ilike: split[0] }},
        { lastName: { $ilike: split[1] }}
      );

      let searchConditions = _.find(query.where.$and, '$or');
      searchConditions.$or.push(searchCondition);
    }
  }]
});
```

PR #74, opened 2017-05-02
[*] Smart Fields - Serialize Smart Fields values for belongsTo association.

pierreco:


routes/resources.js
```diff
@@ -32,6 +32,34 @@ module.exports = function (app, model, Implementation, integrator, opts) {
         } else if (_.isArray(field.type)) {
           record[field.field] = [];
         }
+      } else if(field.reference) {
+        console.log('coucou',field);
```
arnaudbesnier: You should review your own code before asking for review.

routes/resources.js
```diff
@@ -10,17 +10,17 @@ var Schemas = require('../generators/schemas');
 
 module.exports = function (app, model, Implementation, integrator, opts) {
   var modelName = Implementation.getModelName(model);
-  var schema = Schemas.schemas[Implementation.getModelName(model)];
+  var schema = Schemas.schemas[modelName];
```
arnaudbesnier: 👍

routes/resources.js
```diff
@@ -10,17 +10,17 @@ var Schemas = require('../generators/schemas');
...
 
   function injectSmartFields(record) {
     return P.each(schema.fields, function (field) {
       if (!record[field.field]) {
         if (field.get || field.value) {
           if (field.value) {
-            logger.warn('DEPRECATION WARNING: Smart Field value method is ' +
-              'deprecated. Please use get method instead.');
+            logger.warn('DEPRECATION WARNING: In your ' + field.field +
+              ' Collection, Smart Field value method is deprecated. ' +
```
arnaudbesnier: I am not sure that `field.field` is always related to a Collection (fullname is not a Collection for example).

routes/resources.js
```diff
@@ -32,6 +32,34 @@ module.exports = function (app, model, Implementation, integrator, opts) {
...
+              } else {
+                valueSmartField = smartFieldObject.get(record[field.field]);
+              }
+              if (valueSmartField && _.isFunction(valueSmartField.then)) {
+                return valueSmartField.then(function (value) {
+                  record[field.field][smartFieldObject.field] = value;
+                });
+              } else {
+                record[field.field][smartFieldObject.field] = valueSmartField;
+              }
```
arnaudbesnier: The 18 previous lines are duplicated from the existing code. You need to factorise it: define a function that will be used by both cases (`if` and `elsif` cases)

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
...
+      if (smartFieldValue && _.isFunction(smartFieldValue.then)) {
+        return smartFieldValue.then(function (value) {
+          record[smartFieldObject.field] = value;
+        });
+      } else {
+        record[smartFieldObject.field] = smartFieldValue;
+      }
+    } else if (_.isArray(smartFieldObject.type)) {
+      record[smartFieldObject.field] = [];
+    }
```
arnaudbesnier: The `elsif` is not part of the Smart Field getter logic. I am not sur you should use it for the associations.

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
...
-          if (value && _.isFunction(value.then)) {
-            return value.then(function (value) {
-              record[field.field] = value;
-            });
-          } else {
-            record[field.field] = value;
-          }
-        } else if (_.isArray(field.type)) {
-          record[field.field] = [];
+        addSmartFieldToRecord(record, field, modelName);
```
arnaudbesnier: this function may return a Promise, so you should call `return addSmartFieldToRecord...`

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
 
 module.exports = function (app, model, Implementation, integrator, opts) {
   var modelName = Implementation.getModelName(model);
-  var schema = Schemas.schemas[Implementation.getModelName(model)];
+  var schema = Schemas.schemas[modelName];
+
+  function addSmartFieldToRecord(record, smartFieldObject, modelName) {
```
arnaudbesnier: About the function name, this is not exactly what the function does. As you iterate over all fields, it does not necessarily add a Smart Field to the record. It does if the field is a Smart Field.

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
...
+    var smartFieldValue;
+    if(smartFieldObject.get || smartFieldObject.value) {
+      if (smartFieldObject.value) {
+        logger.warn('DEPRECATION WARNING: In your ' + modelName +
+          ' Model, Smart Field value method is deprecated. ' +
+          'Please use get method instead. ');
```
arnaudbesnier: What do you think about my previous comment about the change you made to this log?

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
...
-            return value.then(function (value) {
-              record[field.field] = value;
-            });
-          } else {
-            record[field.field] = value;
-          }
-        } else if (_.isArray(field.type)) {
-          record[field.field] = [];
+        addSmartFieldToRecord(record, field, modelName);
+      } else if(field.reference) {
```
arnaudbesnier: You need to be consistent with your code style: 
`} else if(field.reference) {`
=>
`} else if (field.reference) {`

routes/resources.js
pierreco: Done

routes/resources.js
pierreco: I have replace by "tryToAddSmartFieldToRecord"

routes/resources.js
pierreco: done

routes/resources.js
pierreco: I have replace by "tryToAddSmartFieldToRecord"

routes/resources.js
pierreco: I think it's simpler for the user to have the name of the model for debug

routes/resources.js
```diff
@@ -10,27 +10,44 @@ var Schemas = require('../generators/schemas');
...
-          }
-        } else if (_.isArray(field.type)) {
-          record[field.field] = [];
+        return tryToAddSmartFieldToRecord(record, field, modelName);
+      } else if (field.reference) {
+        // Inject Smart Field in BelongsTo Fields
+        var belongsToModelName = field.reference.split('.')[0];
+        var referenceFieldSchema = Schemas.schemas[belongsToModelName];
+        if (referenceFieldSchema) {
+          _.each(referenceFieldSchema.fields, function (fieldObject) {
```
arnaudbesnier: If the Smart Fields return Promises, the current implementation does not work. I think you should wait for Promises to be resolved before sending the result to the front.

Here is an example of "asynchronous" Smart Fields to debug the current implementation:

```
'use strict';
const _ = require('lodash');
const models = require('../models');
const Liana = require('forest-express-sequelize');

Liana.collection('user', {
  fields: [{
    field: 'fullName',
    type: 'String',
    value: function (user) {
      return models.user.findById(2).then(() => {
        if (user.firstName && user.lastName) {
          return `${user.firstName} ${user.lastName}`;
        } else {
          return null;
        }
      });
    },
    search: function (query, search) {
      var s = models.sequelize;
      var split = search.split(' ');

      var searchCondition = s.and(
        { firstName: { $ilike: split[0] }},
        { lastName: { $ilike: split[1] }}
      );

      let searchConditions = _.find(query.where.$and, '$or');
      searchConditions.$or.push(searchCondition);
    }
  }, {
    field: 'fullName2',
    type: 'String',
    value: function (user) {
      return models.user.findById(2).then(() => {
        if (user.firstName && user.lastName) {
          return `${user.firstName} ${user.lastName}`;
        } else {
          return null;
        }
      });
    },
    search: function (query, search) {
      var s = models.sequelize;
      var split = search.split(' ');

      var searchCondition = s.and(
        { firstName: { $ilike: split[0] }},
        { lastName: { $ilike: split[1] }}
      );

      let searchConditions = _.find(query.where.$and, '$or');
      searchConditions.$or.push(searchCondition);
    }
  }]
});
```

pierreco: Done

PR #76, opened 2017-05-09
modify attributes on virtual field update

vquesnel:
When updating virtual field modify only attributes[field.field] and not set attributes to this virtual field data

### DIFF
--- a/deserializers/resource.js
+++ b/deserializers/resource.js
@@ -30,7 +30,7 @@ function ResourceDeserializer(Implementation, model, params,
               return attributes;
             });
         } else {
-          attributes = field.set(attributes, attributes[field.field]);
+          attributes[field.field] = field.set(attributes, attributes[field.field]);
           return attributes;
         }
       })


arnaudbesnier:
Hi @vquesnel, thanks for the contribution.

I don't really understand the issue.
If you define a Smart Field `set` method, this method is supposed to return the record object once done (see the documentation http://doc.forestadmin.com/developers-guide/#updating-a-smart-field).

Can you explain your use case or send the implementation of your Smart Field `set` method? It would help a lot. 

🌲🌲🌲

vquesnel:
`Liana.collection('OrderItem', {
  fields: [{
    field: '_unitPrice',
    type: 'Number',
    get(item) {
      return item.unitPrice / 100;
    },
    set(item, value) {
      return item.unitPrice = Math.round(value * 100);
    },
  }],
});
`
This is my smart field to calculate the unit price for each item, but when I try to update an item I've got an error telling me that it can't creat property "updatedAt' on number <XXXX> because:

`attributes = field.set(attributes, attributes[field.field]);
  return attributes;`
do not return an object but a value, on the other hand 

`attributes[field.field] = field.set(attributes, attributes[field.field]);
  return attributes;`
return the attributes object with the updated field.

arnaudbesnier:
@vquesnel ,

As you can see in the doc, you have to set your Smart Field this way:
```
Liana.collection('OrderItem', {
  fields: [{
    field: '_unitPrice',
    type: 'Number',
    get(item) { return item.unitPrice / 100; },
    set(item, value) {
      item.unitPrice = Math.round(value * 100);
      // NOTICE: You must return the record here. 
      return item;
    }
  }]
});
```

vquesnel:
Oops, my bad i read a bit to fast your doc 😣
Sorry!

vquesnel:
All work perfectly with this modification! 👍

PR #79, opened 2017-05-15
[*] Smart Fields - Fix some bad Smart Fields getter calls on records list and detail display

arnaudbesnier:


routes/resources.js
```diff
@@ -35,7 +35,8 @@ module.exports = function (app, model, Implementation, integrator, opts) {
         } else if (_.isArray(field.type)) {
           record[field.field] = [];
         }
-      } else if (field.reference) {
```
arnaudbesnier: The second condition prevents the value injection for "hasMany" associations

PR #82, opened 2017-05-30
Improve error logging and provide options to bypass Winston

louisremi:
I've just spent most of my day tracking down another error forest-express was trapping :-)

Usually I go straight to services/error-handler.js and add a `console.error(error)` there, so I have access to the full stack trace.
This time, Forest wasn't displaying any of the smart-fields and smart-segments I had defined in one of my collection in the '/forest' folder, without logging any error to the console, and my usual trick was helpless. It took me quite some time to figure out that requireAllModels was trapping any errors in those files :-(
After that, I searched for a global solution that would allow stack traces to be logged. This proved more difficult than expected, as Winston is a strange beast (that I will personally avoid in libraries I develop): the strings it logs only appear in the terminal, while I only use Chrome's console with Node's `--inspect` flag. I found a solution to log both to Chrome's console and the terminal when in development mode, and log only to the terminal in other cases.

Please let me know if this would work for you :-)

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,26 +20,22 @@ var Integrator = require('./integrations');
 var errorHandler = require('./services/error-handler');
 var codeSyntaxInspector = require('./utils/code-syntax-inspector');
 
-function requireAllModels(Implementation, modelsDir, displayMessage) {
+function requireAllModels(Implementation, modelsDir, isDirRequired) {
   if (modelsDir) {
     return readdirAsync(modelsDir)
       .each(function (file) {
-        try {
-          require(path.join(modelsDir, file));
-        } catch (error) { }
+        require(path.join(modelsDir, file));
       })
       .then(function () {
         return _.values(Implementation.getModels());
       })
       .catch(function (error) {
-        if (displayMessage) {
-          if (error.code === 'ENOENT') {
-            logger.error('Your Forest modelsDir option you configured does not ' +
-              'seem to be an existing directory.');
-          } else {
-            logger.error('Cannot read your models for the following reason: ' +
-              error);
-          }
+        if (error.code === 'ENOENT' && isDirRequired) {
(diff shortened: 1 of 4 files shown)


arnaudbesnier:
hey @louisremi I totally agree about the error logging improvement.
Thanks for pushing us to go in the right direction 👍

arnaudbesnier:
Hi @louisremi I created a new PR (largely inspired by yours) for stack traces display here:
https://github.com/ForestAdmin/forest-express/pull/85

About the error catching, can you explain exactly what was your error. Reading the code, I guess you had an issue with the Forest customization code. But I'd like to understand the issue, because I feel that your fix adds a regression (bad error display) if you don't have a Forest customisation folder `/forest`.

louisremi:
Hey @arnaudbesnier, thank you for taking a look a this PR. Unfortunately, #85 doesn't solve my problems:
1. If you have a syntax error or any other kind of error in a file in the `/forest` dir, this error will be trapped, and all the smart-fields, smart-actions, etc. defined there will be ignored silently :-/
The changes I made to requireAllModels fix that problem, and I was cautious not to mess things up when the `/forest` dir is entirely absent. You can try this for yourself.
2. I need an option to log with console.log instead of Winston: I use Chrome's console instead of the terminal to watch over my running server and debug it, but for some reason Winston is unable to log any message there. The changes I've made to the logger service make this possible.

Could we find a solution for these?

arnaudbesnier:
@louisremi yes I know that your 1st point is not solved by https://github.com/ForestAdmin/forest-express/pull/85. Now that it is clearer, I'll try to reproduce and fix it in a new PR.

About the 2nd point, I understand what you need, but I just need a way to test your usecase, I am not familiar with such configuration. If you help me to learn I'd be happy to help. ;)

PR #97, opened 2017-08-24
Allow users to call a ResourcesRoute from their app.

SeyZ:
## Pull Request checklist:

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,6 +52,7 @@ function requireAllModels(Implementation, modelsDir, displayMessage) {
 
 exports.Schemas = Schemas;
 exports.logger = logger;
+exports.ResourcesRoute = {};
 
 exports.init = function (Implementation) {
   var opts = Implementation.opts;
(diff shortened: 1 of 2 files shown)


SeyZ:
[Oyst want to have a Lumber generated admin with multi databases support](https://trello.com/c/0bb74tZm/1727-oyst-want-to-have-a-lumber-generated-admin-with-multi-databases-support)

SeyZ:
With that feature, I can do this easily:

```
var express = require('express');
var router = express.Router();
var liana = require('forest-express-sequelize');
var Serializer = require('forest-express-sequelize').ResourceSerializer;

function orders(req, res, next) {
  req.query.filterType = 'and';
  req.query.filter = { merchant_id: req.params.merchantId };

  liana.ResourcesRoute.order.list(req, res, next);
}

router.get('/merchant/:merchantId/relationships/orders', orders);

module.exports = router;
```

index.js
```diff
@@ -52,6 +52,7 @@ function requireAllModels(Implementation, modelsDir, displayMessage) {
 
 exports.Schemas = Schemas;
 exports.logger = logger;
+exports.ResourcesRoute = {};
```
arnaudbesnier: `ResourcesRoutes` with a `s` seems more understandable.

arnaudbesnier: `RoutesPerModel` is a better name?
`RoutesPerModel.order.list`?

SeyZ: I agree about RoutesPerModel is more explicit. But I also want to be consistent with the other classes: ResourceSerializer, ResourceDeserializer.

If we change, let's schedule a global name migration with backward compatibility. Honestly, I think we can move forward like this. Tell me what you think.

PR #118, opened 2017-10-20
[*] Smart Fields - Prevent crash when trying to compute Smart Field value.

lucasscariot:
## Pull Request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 ### Changed
 - Smart Relationships - Add a warning if a Smart Collection does not define the "idField" attribute necessary for Smart Relationships. 
 
+### Fixed
+- Smart Fields - Prevent crash when trying to compute Smart Field value.
+
 ## RELEASE 1.3.6 - 2017-10-11
 ### Changed
 - Sessions - Display a clean error message if the renderingId and envSecret are missing or inconsistent.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,7 @@
 ## [Unreleased]
 ### Changed
 - Smart Relationships - Add a warning if a Smart Collection does not define the "idField" attribute necessary for Smart Relationships. 
-
-### Fixed
-- Smart Fields - Prevent crash when trying to compute Smart Field value.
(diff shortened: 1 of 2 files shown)


lucasscariot:
[(1) An error on smart field computation should be catched and display a warning to let the data appear in the list view anyway.](https://trello.com/c/aY6w4Muj/1839-1-an-error-on-smart-field-computation-should-be-catched-and-display-a-warning-to-let-the-data-appear-in-the-list-view-anyway)

services/smart-fields-values-injector.js
```diff
@@ -8,13 +8,18 @@ function SmartFieldsValuesInjector(record, modelName) {
...
     if (field.value) {
       logger.warn('DEPRECATION WARNING: Smart Fields "value" method is ' +
         'deprecated. Please use "get" method in your collection ' +
         modelName + ' instead.');
     }
 
-    var value = field.get ? field.get(record) : field.value(record);
+    try {
+      value = field.get ? field.get(record) : field.value(record);
+    } catch (e) {
```
arnaudbesnier: `e` => `error`

services/smart-fields-values-injector.js
```diff
@@ -8,13 +8,18 @@ function SmartFieldsValuesInjector(record, modelName) {
...
+      logger.error(e);
```
arnaudbesnier: I think you can be more explicit for the developer about what happened. It is good to send the stack trace but an explanation is even better.
Something like:
```javascript
logger.error('Cannot retrieve the ' + field.field + ' value because of an internal error in the getter implementation: ' + error);
```

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 ### Changed
 - Smart Relationships - Add a warning if a Smart Collection does not define the "idField" attribute necessary for Smart Relationships. 
 
+### Fixed
+- Smart Fields - Prevent crash when trying to compute Smart Field value.
```
arnaudbesnier: More an improvement than a "real" fix:

```
### Changed
- Smart Fields - Prevent the Smart Fields computation errors to generate a crash and handle it letting the value empty.
```

PR #157, opened 2018-03-13
[*] Smart Elements - Fix error swallowing on load and clean some useless code

ArnaudValensi:
## Pull Request checklist:


index.js
```diff
@@ -171,22 +160,22 @@ exports.init = function (Implementation, dependencies) {
 
           if (opts.configDir) {
             directorySmartImplementation = path.resolve('.', opts.configDir);
+
+            if (!fs.existsSync(directorySmartImplementation)) {
+              logger.error('Your Forest modelsDir option you configured does not seem to be an existing directory.');
+              return P.resolve([]);
```
arnaudbesnier: You don't need to declare a Promise here:
`return [];` is enough.

index.js
```diff
@@ -171,22 +160,22 @@ exports.init = function (Implementation, dependencies) {
...
-                return requireAllModels(Implementation,
-                  directorySmartImplementation, false);
-              }
-            });
+            if (fs.existsSync(directorySmartImplementation)) {
+              return requireAllModels(Implementation, directorySmartImplementation);
+            } else {
+              return P.resolve([]);
+            }
+          }
```
arnaudbesnier: Looks like you duplicated the logic.
the `directorySmartImplementation` value and the error log are the only things that are specific to `configDir` presence.

```javascript
var directorySmartImplementation = path.resolve('.', opts.configDir || '/forest');

if (fs.existsSync(directorySmartImplementation)) {
  return requireAllModels(Implementation, directorySmartImplementation);
} else {
  if (opts.configDir) {
    logger.error('The Forest modelsDir option you configured does not seem to be an existing directory.');
  }
  return [];
}
```
Does the same job, with less code/logic.

ArnaudValensi: `path.resolve('.') + '/forest'` is not the same as `path.resolve('.', '/forest')`

arnaudbesnier: then `path.resolve('.', 'forest')` is equivalent to `path.resolve('.') + '/forest'`

PR #190, opened 2018-07-09
[+] Authentication - Add two factor authentication using time-based one-time password

ArnaudValensi:
## Pull Request checklist:


ArnaudValensi:
[(?) 2FA - Front/Express/Back - Registration](https://trello.com/c/FunJ61p9/2683-2fa-front-express-back-registration)

.babelrc
```diff
@@ -1,3 +1,12 @@
 {
-  "presets": ["env"]
+  "presets": ["env"],
+  "plugins": [
+    [
+      "transform-runtime",
```
ArnaudValensi: This is needed to make the async/await work when transpiled.

.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     },
     'extends': 'eslint:recommended',
     'parserOptions': {
-        'ecmaVersion': 6
+        'ecmaVersion': 2017
```
ArnaudValensi: async/await come since es2017,

package.json
```diff
@@ -45,6 +47,7 @@
   "devDependencies": {
     "babel-cli": "6.26.0",
     "babel-core": "6.26.0",
+    "babel-plugin-transform-runtime": "6.23.0",
```
ArnaudValensi: This is needed to make the async/await work when transpiled.

src/routes/sessions.js
```diff
@@ -62,50 +60,76 @@ module.exports = function (app, opts, dependencies) {
     };
   }
 
+  function getTwoFactorResponse(user) {
+    if (process.env.LIANA_2FA_SECRET  === undefined) {
```
arnaudbesnier: Double spaces before `===`

arnaudbesnier: So we need a specific `LIANA_2FA_SECRET` environment variable? Why?
`LIANA_` is not the conventional prefix anyway for Forest environment variables.

src/routes/sessions.js
```diff
@@ -62,50 +60,76 @@ module.exports = function (app, opts, dependencies) {
...
     const envSecret = opts.envSecret;
+    const { email, password } = request.body;
 
-    return ipWhitelist.retrieve(envSecret)
-      .then(function () { return new AllowedUsersFinder(renderingId, envSecret).perform(); })
-      .then(function (allowedUsers) {
-        if (allowedUsers.length === 0) {
-          throw new Error(errorMessages.SESSION.NO_USERS);
-        }
+    (async function asyncWrapper() {
```
arnaudbesnier: What is the benefit of such code VS `try/catch` syntax?

test/routes/sessions.js
```diff
@@ -0,0 +1,291 @@
...
+            expect(token).to.be.undefined;
+            expect(twoFactorAuthenticationEnabled).to.be.true;
+            expect(userSecret).to.equal(expectedUserSecret);
+
+            done();
+          });
+      });
+    });
+
+    describe('with 2FA enabled but not configured', () => {
```
ArnaudValensi: I just find a type, it is: `with 2FA enabled AND configured`.

src/routes/sessions.js
```diff
@@ -62,50 +60,76 @@ module.exports = function (app, opts, dependencies) {
...
 
-        return bcrypt.compare(request.body.password, user.password)
-          .then(function (isEqual) {
-            if (!isEqual) { throw new Error(); }
-            return user;
-          });
-      })
-      .then(sendToken(response, renderingId))
-      .catch(formatAndSendError(response));
+      await sendToken(response, renderingId)(user);
```
arnaudbesnier: Maybe you can simplify the `sendToken` implementation without having to create a function in the function body.
(`sendToken(response, renderingId, user)`)

src/services/authorization-finder.js
```diff
@@ -0,0 +1,31 @@
...
+function AuthorizationFinder(renderingId, email, password, envSecret) {
+  this.perform = function () {
+    return new P(function (resolve, reject) {
+      const forestUrl = new ServiceUrlGetter().perform();
+
+      request
+        .get(`${forestUrl}/liana/v1/renderings/${renderingId}/authorization`)
+        .set('forest-secret-key', envSecret)
+        .set('email', email)
+        .set('password', password)
```
arnaudbesnier: Quite strange to send this in the headers, why don't you do a `POST`?

src/routes/sessions.js
```diff
@@ -62,50 +60,76 @@ module.exports = function (app, opts, dependencies) {
...
   function loginWithPassword(request, response) {
     const renderingId = request.body.renderingId;
     const envSecret = opts.envSecret;
+    const { email, password } = request.body;
 
-    return ipWhitelist.retrieve(envSecret)
-      .then(function () { return new AllowedUsersFinder(renderingId, envSecret).perform(); })
-      .then(function (allowedUsers) {
-        if (allowedUsers.length === 0) {
-          throw new Error(errorMessages.SESSION.NO_USERS);
```
arnaudbesnier: If not used (`NO_USERS`) this should be cleaned.

src/services/server-response-handler.js
```diff
@@ -0,0 +1,26 @@
...
+const P = require('bluebird');
+
+function ServerResponseHandler(error, result) {
+  this.perform = () => {
+    return P.try(() => {
+      if (result.status === 200 && result.body && result.body.data && result.body.data.attributes) {
+        return result.body.data;
+      } else {
+        if (result.status === 0) {
+          logger.error(errorMessages.SESSION.SERVER_DOWN);
```
arnaudbesnier: If it is the `server-response-handler` the error messages are related to sessions.

src/routes/sessions.js
ArnaudValensi: With try/catch you still need to wrap the `await` call to an `async function ...`

src/routes/sessions.js
ArnaudValensi: The 2FA user secret sent to the client is the combination of the partial user secret of the server and the `LIANA_2FA_SECRET`. So that, we, at forest, cannot have access to the 2FA user secret.

src/services/authorization-finder.js
ArnaudValensi: Because on a rest perspective, we want to retrieve (GET) not create (POST). And I don't want to pass the cresentials as query params to avoid leaks in logs. The last viable alternative to me was headers.

src/services/user-secret-creator.js
```diff
@@ -0,0 +1,20 @@
+const base32Encode = require('base32-encode');
+const xor = require('bitwise-xor');
+
+// NOTICE: This service combines the 2FA secret stored on the forest server to the local secret
+//         salt. This guarantees that only the owner of the server and the concerned end user can
+//         know the final key.
+//         This is done by using a bitwise exclusive or operation, which guarantees the key to stay
+//         unique, so it is impossible for two users to have the same key.
```
arnaudbesnier: Ok!

(Considering that the 2FSecretSalt is the same for all users of a project and that each user secret is supposed to be different accross the list of users of a project.)

src/services/login-handler.js
```diff
@@ -0,0 +1,89 @@
...
+  dependencies
+}) {
+  if (dependencies.GoogleAuthorizationFinder) {
+    GoogleAuthorizationFinder = dependencies.GoogleAuthorizationFinder;
+  }
+
+  const { forestToken, email, password } = authData;
+
+  function getTwoFactorResponse(user) {
+    const twoFactorSecretSalt = process.env['2FA_SECRET_SALT'];
```
arnaudbesnier: Capital letters for constants makes it easier to see that this variable is "special".

arnaudbesnier: `2FA_SECRET_SALT` to prefix with `FOREST`

PR #196, opened 2018-09-04
[-] Authentication - Improve the log message when 2FA secret key is not set

ArnaudValensi:
## Pull Request checklist:

### DIFF
--- a/src/services/login-handler.js
+++ b/src/services/login-handler.js
@@ -35,14 +35,12 @@ function LoginHandler({
     const TWO_FACTOR_SECRET_SALT = process.env.FOREST_2FA_SECRET_SALT;
 
     if (TWO_FACTOR_SECRET_SALT === undefined) {
-      logger.error('Cannot use the two factor authentication because the environment variable "FOREST_2FA_SECRET_SALT" is not set.');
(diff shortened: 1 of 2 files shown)


ArnaudValensi:
[(8) 2FA - Rails - Registration & Login](https://trello.com/c/mYrGbDPe/2691-8-2fa-rails-registration-login)

src/services/login-handler.js
```diff
@@ -36,6 +36,7 @@ function LoginHandler({
 
     if (TWO_FACTOR_SECRET_SALT === undefined) {
       logger.error('Cannot use the two factor authentication because the environment variable "FOREST_2FA_SECRET_SALT" is not set.');
+      logger.error('You can generate it using this command: `$ openssl rand -hex 10`');
```
arnaudbesnier: IMO this info should be in the doc or in the feature onboarding steps not in an error log.
That's why I removed it in the first implementation.
Once a doc section has been created, we could had the link in this error message.

If you really want to add this, I would concatenate the new sentence to the first error log message to make it obvious that they are related.
(In the case a third error message appears for example)

PR #203, opened 2018-10-11
Fix/nested relationship smart view

VincentMolinie:
## Pull Request checklist:

- [X] Write an explicit title for the Pull Request
- [X] Write changes made in the CHANGELOG.md
- [X] Wonder if you can improve the existing code

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "express-jwt": "5.3.1",
     "inflected": "2.0.4",
     "ip-utils": "git+https://github.com/ForestAdmin/ip-utils.git#5f88562ba53fedcdc0374937fca0fdb71fa4923c",
-    "jsonapi-serializer": "3.4.1",
+    "jsonapi-serializer": "3.6.0",
     "jsonwebtoken": "8.1.1",
     "lodash": "4.10.0",
     "moment": "2.19.4",
--- a/src/serializers/resource.js
+++ b/src/serializers/resource.js
@@ -42,7 +42,34 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
   this.perform = function () {
     var typeForAttributes = {};
 
-    function getAttributesFor(dest, fields) {
+    function getRelatedLinkForBelongsTo(modelName, idField) {
+      return function (record, current) {
+        return `/forest/${modelName}/${current[idField]}`;
+      };
+    }
+
+    function getRelatedLinkForHasMany(modelName, relationshipName, idField) {
+      return function (record, current, parent) {
+        return `/forest/${modelName}/${parent[idField]}/relationships/${relationshipName}`;
+      };
+    }
+
+    function getReferenceField(referenceSchema, field) {
+      let fieldReference = referenceSchema.idField;
+
+      if (_.isArray(field.type) && !fieldReference && referenceSchema.isVirtual) {
+        if (_.find(referenceSchema.fields, function (field) { return field.field === 'id'; })) {
+          fieldReference = 'id';
+        } else {
+          logger.warn('Cannot find the \'idField\' attribute in your \'' +
+            referenceSchema.name + '\' Smart Collection declaration.');
+        }
+      }
+
+      return fieldReference || 'id';
+    }
+
+    function getAttributesFor(dest, fields, include = true) {
       _.map(fields, function (field) {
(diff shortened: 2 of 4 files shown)


VincentMolinie:
[P2 - Plugify can't get his smart view to work properly](https://trello.com/c/zdhDFnIZ/2937-p2-plugify-cant-get-his-smart-view-to-work-properly)

VincentMolinie:
To add more context about how JSON API works:
First thing we need to know is that if in a relationships, there is a link specified even if the data are retrieved in the included, Ember Data will do a GET request on the url. So link must be here ONLY if the data is not included. Then where there is data in relationships { data: { id: '1', type: 'student'}} and a link the data will always be the first one to be called. If none of them is specified, the relation is considered as null.

package.json
```diff
@@ -35,7 +35,7 @@
     "express-jwt": "5.3.1",
     "inflected": "2.0.4",
     "ip-utils": "git+https://github.com/ForestAdmin/ip-utils.git#5f88562ba53fedcdc0374937fca0fdb71fa4923c",
-    "jsonapi-serializer": "3.4.1",
+    "jsonapi-serializer": "3.6.1",
```
VincentMolinie: This fix some issue with the serializer and allow to return null from relationshipLinks function and not include the link in this case

src/serializers/resource.js
```diff
@@ -42,7 +42,31 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
   this.perform = function () {
     var typeForAttributes = {};
 
-    function getAttributesFor(dest, fields) {
+    function getRelatedLinkForHasMany(modelName, relationshipName) {
+      return function (record, current, parent) {
+        if (current && current[relationshipName]) {
```
VincentMolinie: We do not want the link to be set if the relation is already included in the payload

src/serializers/resource.js
```diff
@@ -73,38 +97,31 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
...
               detectFieldWithSpecialFormat(field, fieldName);
             });
 
+            const modelName = Implementation.getModelName(model);
+            const relatedFunction = field.relationship === 'BelongsTo'
+              ? null
+              : getRelatedLinkForHasMany(modelName, field.field);
             dest[fieldName] = {
               ref: fieldReference,
+              nullIfMissing: true,
```
VincentMolinie: We need this or the links will not be included if the relation is not in the records sent to the serializer

src/serializers/resource.js
```diff
@@ -73,38 +97,31 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
...
-                }
-              }
+                related: relatedFunction,
+              },
             };
 
-            if (_.isArray(field.type)) {
-              dest[fieldName].ignoreRelationshipData = true;
+            if (_.isArray(field.type) || !include) {
+              dest[fieldName].ignoreRelationshipData = include;
```
VincentMolinie: We do not include from second level (We never include relations of a relation)

src/serializers/resource.js
```diff
@@ -141,7 +158,35 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
       });
     }
 
-    var serializationOptions = {
+    function defineRelationshipId(records, fields, goDeeper = true) {
```
VincentMolinie: Recursive function to initialize belongsTo with an id if not set (based on foreignKey) so the serializer can return the `{ data: { id: 2 } }`

src/serializers/resource.js
```diff
@@ -141,7 +158,35 @@ function ResourceSerializer(Implementation, model, records, integrator, opts, me
...
+      recordArray.forEach((record) => {
+        fields.forEach((field) => {
+          if (!field.reference) {
+            return;
+          }
+
+          const fieldName = field.field;
+          const [referenceType,referenceKey] = field.reference.split('.');
+          const referenceSchema = Schemas.schemas[referenceType];
+          if (goDeeper && record[fieldName]) {
```
VincentMolinie: We only do it on one level. But we could do it recursively like on forest-rails (@arnaudbesnier Your call)

arnaudbesnier:
Let's keep this for later. We need stability now and I don't want to introduce unecessary feature. Especially on the liana side.

PR #214, opened 2019-04-15
Add an option onlyCrudModule to expose only the services without the …

SeyZ:
…Forest Admin's init

## Pull Request checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -80,10 +80,25 @@ exports.ensureAuthenticated = (request, response, next) => {
 
 let alreadyInitialized = false;
 
+function buildSchema(Implementation) {
+  const { opts } = Implementation;
+
+  const absModelDirs = opts.modelsDir ? path.resolve('.', opts.modelsDir) : undefined;
+  return requireAllModels(Implementation, absModelDirs)
+    .then((models) => {
+      const integrator = new Integrator(opts, Implementation);
+      return Schemas.perform(Implementation, integrator, models, opts)
+        .thenReturn([models, integrator]);
+    });
+}
+
 exports.init = (Implementation) => {
   const { opts } = Implementation;
+  if (opts.onlyCrudModule === true) {
+    return buildSchema(Implementation).spread(models => models);
+  }
+
   const app = express();
-  let integrator;
 
   if (alreadyInitialized) {
     logger.warn('Forest init function called more than once. Only the first call has been processed.');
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -80,8 +80,23 @@ exports.ensureAuthenticated = (request, response, next) => {
 
 let alreadyInitialized = false;
 
+exports.initServices = (Implementation) => {
+  const { opts } = Implementation;
+  const integrator = new Integrator(opts, Implementation);
```
arnaudbesnier: Why isn't it done after the `requireAllModels` like at line 184?

src/index.js
```diff
@@ -80,8 +80,23 @@ exports.ensureAuthenticated = (request, response, next) => {
 
 let alreadyInitialized = false;
 
+exports.initServices = (Implementation) => {
```
arnaudbesnier: Is it necessary to expose (`exports`) this function?

src/index.js
```diff
@@ -80,8 +80,23 @@ exports.ensureAuthenticated = (request, response, next) => {
...
+  const { opts } = Implementation;
+  const integrator = new Integrator(opts, Implementation);
+
+  // Init
+  const absModelDirs = opts.modelsDir ? path.resolve('.', opts.modelsDir) : undefined;
+  requireAllModels(Implementation, absModelDirs)
+    .then(models => Schemas.perform(Implementation, integrator, models, opts));
+};
```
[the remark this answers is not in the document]
SeyZ: fixed

src/index.js
SeyZ: fixed

src/index.js
SeyZ: I have now use the same code than before expect it's correctly encapsulated. Didn't change the order of functions

src/index.js
```diff
@@ -163,30 +180,23 @@ exports.init = (Implementation) => {
...
 
-          if (fs.existsSync(directorySmartImplementation)) {
-            return requireAllModels(Implementation, directorySmartImplementation);
-          }
-          if (opts.configDir) {
-            logger.error('The Forest modelsDir option you configured does not seem to be an existing directory.');
-          }
-          return [];
-        })
-        .thenReturn(models);
```
arnaudbesnier: Here, in the old code, it returns models from `absModelDirs`.
And in the new code it may return an empty array or models detected in the Smart directory.
The failing test might be a symptom of that.

src/index.js
```diff
@@ -163,30 +180,23 @@ exports.init = (Implementation) => {
...
+      if (fs.existsSync(directorySmartImplementation)) {
+        return requireAllModels(Implementation, directorySmartImplementation);
+      }
+      if (opts.configDir) {
+        logger.error('The Forest modelsDir option you configured does not seem to be an existing directory.');
+      }
+      return [];
```
arnaudbesnier: Why don't we also execute this part of the logic with `onlyCrudModule: true` to load Collections Smart configuration?

src/index.js
```diff
@@ -80,8 +80,25 @@ exports.ensureAuthenticated = (request, response, next) => {
...
+  const { opts } = Implementation;
+
+  const absModelDirs = opts.modelsDir ? path.resolve('.', opts.modelsDir) : undefined;
+  return requireAllModels(Implementation, absModelDirs)
+    .then((models) => {
+      const integrator = new Integrator(opts, Implementation);
+      return Schemas.perform(Implementation, integrator, models, opts)
+        .thenReturn(models);
+    });
+}
```
arnaudbesnier: Looks good 👍

src/index.js
```diff
@@ -80,10 +80,25 @@ exports.ensureAuthenticated = (request, response, next) => {
...
+      const integrator = new Integrator(opts, Implementation);
+      return Schemas.perform(Implementation, integrator, models, opts)
+        .thenReturn([models, integrator]);
+    });
+}
+
 exports.init = (Implementation) => {
   const { opts } = Implementation;
+  if (opts.onlyCrudModule === true) {
+    return buildSchema(Implementation).spread(models => models);
```
arnaudbesnier: Is it consistent to return `models` for this kind of initialization (with `.spread(models => models);`)?
Do you really need this?

It will return an `app` object in the default initialization that why I am asking.

PR #238, opened 2019-06-11
[-] Technical - Makes the JWT lighter and consistent across lianas

larcin:
## Pull Request checklist:


larcin:
[Authentication - JWT object is different between express and rails](https://trello.com/c/wfWMaGbv/3482-authentication-jwt-object-is-different-between-express-and-rails)

test/routes/sessions.js
```diff
@@ -178,23 +178,37 @@ describe('API > Sessions', () => {
...
+              first_name: 'user',
+              last_name: 'last',
+              teams: ['Operations'],
+            });
+
+            done();
+          });
+      });
+
+      it('should return a valid jwt (new format)', (done) => {
```
arnaudbesnier: [Cleaning] Isn't it a duplicate test? The one above looks very very similar.
So I am pretty sure this one is useless.

src/services/login-handler.js
```diff
@@ -54,24 +54,14 @@ function LoginHandler({
...
-            id: renderingId,
-          }],
-        },
-      },
+      email: user.email,
+      first_name: user.first_name,
+      last_name: user.last_name,
+      teams: user.teams,
+      renderingId,
     }, authSecret, {
```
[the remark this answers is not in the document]
larcin: I don't get why there should be a single team. Can't the users have several teams ?

src/middlewares/permissions.js
```diff
@@ -24,6 +27,9 @@ function createCheckPermission(environmentSecret, collectionName) {
     const renderingId = getRenderingIdFromUser(request.user);
     const permissionName = searchToEdit ? 'searchToEdit' : 'list';
 
+    // NOTICE: Unauthorize user if he's using a deprecated jwt
+    if (!renderingId) next(httpError(401, 'Deprecated JWT'));
+
```
arnaudbesnier: [Code] For maintenance, we shouldn't have code duplication between lines 12-14 and 30-32.

I don't know if you move the logic in it already, but isn't it the role of  `ensureAuthenticated` function in `auth.js` to respond 401 if the token has the old format.

You could implement something like:
```javascript
function ensureAuthenticated(request, response, next) {
  if (!request.user) {
    return next(new error.Unauthorized(ERROR_MESSAGE_DEFAULT));
  }
  
  // NOTICE: Automatically logout users trying to access the API with a token having an old data format.
  if (request.user.type) {
    return next(new error.Unauthorized(ERROR_MESSAGE_TOKEN_OLD));
  }
  
  return next();
}
```

PR #275, opened 2019-11-07
[*] Smart Relationships - Smart Relashionships returns referenced Smart Fields values

rap2hpoutre:
## Description

Currently, smart fields are not computed when loaded through a smart relationship (see associated Trello ticket). This PR computes (recursively) the smart fields of the associated model.

### Caveats:

If the referenced smart field (via the smart relationship) does heavy work (async call), it will slow down the whole process. In other words, currently, the smart fields of the related models are ignored, now they are computed.

## Pull Request checklist:

### Bonus

- [x] Add tests!


rap2hpoutre:
[P3 - User can not see their reference field when it's a smartfield on refresh collection page](https://trello.com/c/njxu6lOk/3845-p3-user-can-not-see-their-reference-field-when-its-a-smartfield-on-refresh-collection-page)

src/services/smart-fields-values-injector.js
```diff
@@ -33,6 +33,19 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
...
+              //         then we consider we have to inject smart fields within the reference.
+              if (result && result.dataValues && field.reference) {
+                const [referenceModelName] = field.reference.split('.');
+                const smartFieldsValuesInjector = new SmartFieldsValuesInjector(
+                  result,
+                  referenceModelName,
+                  fieldsPerModel,
+                );
+                smartFieldsValuesInjector.perform();
+              }
```
arnaudbesnier: [Consistency] You'll also have to use/implement this logic at line 56.
Here is the implementation case of Smart Field `getter` that return a promise (`_.isFunction(value.then)`).

You'll also have to handle this for Smart Field `getter` that return a value IMO.

src/services/smart-fields-values-injector.js
```diff
@@ -33,6 +33,19 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
...
             .then((result) => {
+              // NOTICE: We need some recursion in order to consider smart fields in smart
+              //         relations. So if the result contains data and the field has a reference
+              //         then we consider we have to inject smart fields within the reference.
+              if (result && result.dataValues && field.reference) {
+                const [referenceModelName] = field.reference.split('.');
+                const smartFieldsValuesInjector = new SmartFieldsValuesInjector(
+                  result,
+                  referenceModelName,
+                  fieldsPerModel,
```
arnaudbesnier: Not really sure what is contained in `fieldsPerModel`.
I am sure you fix works as you requested a review, but I wonder if this will always contain the necessary context.

src/services/smart-fields-values-injector.js
```diff
@@ -33,6 +33,19 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
         if (_.isFunction(value.then)) {
           return value
             .then((result) => {
+              // NOTICE: We need some recursion in order to consider smart fields in smart
```
arnaudbesnier: [Food for thought] Recursivity is powerful but could also increase the response time of the request.
Actually whatever the way it is fixed it will increase the response time.

For instance, if you have a Smart Relationship with asynchronous requests, and a pagination of 20 records:
- with the current code base you will have 20 asynchronous calls
- with this code, you'll have 20 asynchronous calls x number of Smart Fields declared in the referenced model.

src/services/smart-fields-values-injector.js
rap2hpoutre: Good catch! Fixed!

src/services/smart-fields-values-injector.js
rap2hpoutre: Yes, sadly 😢

```
number_of_records_paginated * number_of_smart_fields_that_returns_a_collection
```

I don't know how to avoid this if we want to consider smart fields in smart relations. We have to keep this caveat in mind though.

src/services/smart-fields-values-injector.js
rap2hpoutre: `fieldsPerModel` contains all the requested fields. Exemple:

```
// Calling this on a lumber instance:
http://example.org/forest/collection_2?fields[col_1]=CHID_and_status&fields[collection_2]=aml_status,created_at,deleted_at,col_1,&timezone=Europe/Paris
```

```
// Gives:
{
  col_1: ['CHID_and_status'],
  collection_2: ['aml_status', 'created_at', deleted_at', 'col_1']
}
```

rap2hpoutre:
@arnaudbesnier Thank you for your review. I consider it ready for a _complete_ review (I added some tests), let me know if it's OK! I hope I did not missed some edge cases...

test/services/smart-fields-values-injector.js
```diff
@@ -0,0 +1,44 @@
...
+  it('should work with a simple smart field', async () => {
+    Schemas.schemas = { users: usersSchema };
+    const record = { id: 123 };
+    const fieldsPerModel = { users: ['id', 'smart'] };
+    const injector = new SmartFieldsValuesInjector(record, 'users', fieldsPerModel);
+    await injector.perform();
+    expect(record).to.be.deep.equal({ id: 123, smart: { foo: 'bar' } });
+  });
+
+  it('should work without smart fields', async () => {
```
arnaudbesnier: [Code Style] More explicit test
```suggestion
    describe('without Smart Fields', () => {
      it('should not modify the record', async () => {
```

arnaudbesnier: I would move it as the first test to ease the test cases progression/readability.

test/services/smart-fields-values-injector.js
```diff
@@ -0,0 +1,44 @@
...
+    usersSchemaWithoutSmartField.fields.shift();
+    Schemas.schemas = { users: usersSchemaWithoutSmartField };
+    const record = { id: 123 };
+    const fieldsPerModel = { users: ['id'] };
+    const injector = new SmartFieldsValuesInjector(record, 'users', fieldsPerModel);
+    await injector.perform();
+    expect(record).to.be.deep.equal({ id: 123 });
+  });
+
+  it('should work with a smart relationship that reference a smart field', async () => {
```
arnaudbesnier: [Code Style] More explicit test
```suggestion
    describe('with a Smart Relationship that reference a collection having a Smart Field', () => {
      it('should inject the Smart Relationship reference', async () => {
        // ...
      }
      
      it('should inject the Smart Field of the record referenced by the Smart Relationship', async () => {
        // ...
      }
```

test/services/smart-fields-values-injector.js
```diff
@@ -0,0 +1,44 @@
...
+const _ = require('lodash');
+const SmartFieldsValuesInjector = require('../../src/services/smart-fields-values-injector');
+const Schemas = require('../../src/generators/schemas');
+const usersSchema = require('../fixtures/users-schema.js');
+const addressesSchema = require('../fixtures/addresses-schema.js');
+
+const { expect } = chai;
+chai.use(chaiSubset);
+
+describe('Service > Smart Fields Values Injector', () => {
```
arnaudbesnier: 🎖Really cool to add tests!

test/fixtures/users-schema.js
```diff
@@ -0,0 +1,127 @@
...
+      isVirtual: false,
+      description: null,
+      reference: null,
+      inverseOf: null,
+      relationships: null,
+      enums: null,
+      validations: [],
+      integration: null,
+    },
+  ],
```
arnaudbesnier: [Code Style] The lighter the fixtures the easier we read and maintain it.
I would reduce the number of declared fields.

All the fields that are not necessary for the test purposes should be deleted IMO,

src/services/smart-fields-values-injector.js
```diff
@@ -17,6 +17,25 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
       }
     }
 
+    // NOTICE: Update the record with smart field value (contains side effects).
```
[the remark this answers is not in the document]
rap2hpoutre: Good catch, thank you!  I created this function to avoid duplication, but it resulted in something weird!

So, based on your comment, I _removed_ this procedural function that performed a sequence of actions.  It result in a more concise code (`+38 / -45`), easier to read _in my opinion_ (considering the Smart Field value as a promise, I can now read the whole code vertically from top to bottom: it actually work well with my mind ; I hope it will work for you too!).

src/services/smart-fields-values-injector.js
```diff
@@ -17,6 +17,25 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
...
+        // NOTICE: Field reference format: `modelName.property`.
+        const [referenceModelName] = field.reference.split('.');
+        const smartFieldsValuesInjector = new SmartFieldsValuesInjector(
+          smartFieldValue,
+          referenceModelName,
+          fieldsPerModel,
+        );
+        await smartFieldsValuesInjector.perform();
+      }
+      // NOTICE: (side effects) Update the record, then update fieldsForHighlightedSearch.
```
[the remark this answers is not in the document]
rap2hpoutre: I removed every comments about side effects (see my previous comment). 

> If you think that addFieldForHighlightIfCandidate method name is not explicit enough, feel free to improve it 

I agree, it's not explicit. I removed it and replaced it by a comment:

```javascript
// NOTICE: String fields can be highlighted.
if (field.type === 'String') {
  fieldsForHighlightedSearch.push(field.field);
}
```

_Side note: there are still side effects in this function, I don't think I should to refactor it since I want to focus on the main issue._

src/services/smart-fields-values-injector.js
```diff
@@ -17,6 +17,25 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
       }
     }
 
+    // NOTICE: Update the record with smart field value (contains side effects).
+    async function updateRecord(smartFieldValue) {
+      // NOTICE: We need some recursion in order to consider smart fields in smart relations.
+      //         So if the smartFieldValue contains data and the field has a reference then we
+      //         consider we have to inject smart fields within the reference.
```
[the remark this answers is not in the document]
rap2hpoutre: Thank you, it's way better

src/services/smart-fields-values-injector.js
```diff
@@ -17,6 +17,25 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel) {
...
+      if (smartFieldValue && smartFieldValue.dataValues && field.reference) {
+        // NOTICE: Field reference format: `modelName.property`.
```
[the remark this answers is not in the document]
rap2hpoutre: Nice. I created a function and moved the comment!

test/fixtures/users-schema.js
rap2hpoutre: :+1: I removed the `updatedAt`, `createdAt` and `deletedAt`

rap2hpoutre:
@arnaudbesnier Thank you for your review 🍻 

2 [review](https://github.com/ForestAdmin/forest-express/pull/275/files/e871a90146896a13a2534d8b237177eaed5b1bbf#diff-6986382034a8928f3bf7c5e28f0ab31c) [comments](https://github.com/ForestAdmin/forest-express/pull/275/files/e871a90146896a13a2534d8b237177eaed5b1bbf#diff-6986382034a8928f3bf7c5e28f0ab31c) are still open (I would like to be sure you agree with the design I implemented); I marked the others as resolved: I took your suggestions! 🙏

src/services/smart-fields-values-injector.js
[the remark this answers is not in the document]
arnaudbesnier: Really not a fan of 1 comment above each line of code. Like this:
```javascript
// NOTICE: Update the record with the the Smart Field value.
record[field.field] = smartFieldValue;
// NOTICE: String fields can be highlighted.
if (field.type === 'String') {
  fieldsForHighlightedSearch.push(field.field);
}
```
This is pretty limited so it does not hurt.
But comments are like code, you have to maintain them (and if you don't do it well, people that read them could be lost).
For instance, I am not sure we want to maintain:
```javascript
// NOTICE: Update the record with the the Smart Field value.
```
comment, as the intension described is really close to the action:
```javascript
record[field.field] = smartFieldValue;
```

PR #302, opened 2019-11-29
feat: add semantic release

rap2hpoutre:
- [x] It uses https://github.com/semantic-release/semantic-release (which includes semantic changelog and follow semantic commit spec).
- [x] We do not need `@forestadmin/devops` anymore
  - ~~[ ] So it could be removed~~ it has not been removed because update coverage needs it.
- [x] For each push to master, the release process starts (update version, changelog, deploy)
  - [x] We need a way to merge dev into master without squashing. It could be done via PR (we have to allow back rebase+merge) or via command line (merge devel in master means I want to release).
  - [x] No more manual change of changelog nor manual release.
- [x] This branch runs a dry run
  - [x] Remove the dry-run before merging
- ~~[ ] Send notification on Slack (needs an App) => maybe use the current system POSPONED~~
- [x] It currently uses my github token and npm token
  - [x] Use another token
- [x] Verify that the commit message is well formed

See: https://travis-ci.org/ForestAdmin/forest-express/jobs/618651403?utm_medium=notification&utm_source=github_status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ jobs:
         on:
           branch: "improvement/semantic-release"
         script:
+          - yarn build
           - npx semantic-release --dry-run --branch "improvement/semantic-release"
     - stage: publish
       if: branch IN ("master")
@@ -34,4 +35,5 @@ jobs:
         on:
           branch: master
         script:
+          - yarn build
           - npx semantic-release --dry-run
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ jobs:
         on:
           branch: "improvement/semantic-release"
         script:
-          - yarn build
           - npx semantic-release --dry-run --branch "improvement/semantic-release"
     - stage: publish
       if: branch IN ("master")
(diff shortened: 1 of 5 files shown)


rap2hpoutre:
Almost ready. Todo:
 - Create a BOT user for `NPM` and `GITHUB` (and use their credentials in CI).
 - Allow back rebase+merge.
 - Remove code about this branch

arnaudbesnier:
I post my question here. I am sure the response are easily accessible but I just want to track the question.
How do we handle beta versions if every release versions increment is automatically done based on the commits?

rap2hpoutre:
@arnaudbesnier It [seems](https://semantic-release.gitbook.io/semantic-release/support/faq#can-i-manually-trigger-the-release-of-a-specific-version) semantic-release does not handle beta-version. We would have to build it ourselves (maybe just `build` then `publish` is enough (and maybe `tag` too)). The problem will be that our betas will be untracked and have an inaccurate changelog. I do not have a strong opinion about this caveat, what's yours?

PR #306, opened 2019-12-03
[+] Initialisation - Make liana call `opts.callback` function after i…

slimee:
…nitialisation if provided.

## Pull Request checklist:

- [X] Write an explicit title for the Pull Request
- [X] Write changes made in the CHANGELOG.md
- [X] Wonder if you can improve the existing code

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    'jest/no-test-return-statement': 0,
     'implicit-arrow-linebreak': 0,
     'no-param-reassign': 0,
     'no-underscore-dangle': 0,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Initialisation - Make liana initialisation use `opts.callback` if provided.
 
 ## RELEASE 5.4.0 - 2019-11-29
 ### Added
--- a/src/index.js
+++ b/src/index.js
@@ -302,13 +302,20 @@ exports.init = (Implementation) => {
       .retrieve(opts.envSecret)
       // NOTICE: An error log (done by the service) is enough in case of retrieval error.
       .catch(() => {}))
+    .then(() => {
+      if (opts.expressParentApp) {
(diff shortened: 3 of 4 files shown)


slimee:
It's sad here: we can't add more tests since the init function is stateful.

slimee:
[Liana - Ensure that the express lianas init function return a promise that once resolved should ensure that everything has been properly setup for the Admin API.](https://trello.com/c/4xk5Jd73/3899-liana-ensure-that-the-express-lianas-init-function-return-a-promise-that-once-resolved-should-ensure-that-everything-has-been-pr)

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    'jest/no-test-return-statement': 0,
```
arnaudbesnier: [Conventions] Do we really want to deactivate this rule for all our tests?

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Initialisation - Make liana initialisation use `opts.callback` if provided.
```
arnaudbesnier: [Dev Experience] Can we purpose a new major release that would expose a promise (vs the callback implementation)?
It would probably be much more "modern" than a callback.

If we release this new liana version as a v6, Lumber will still use liana v5 unless we implement the new version using a liana v6. It should work this way, right?

slimee: Yes, we can do this.

.eslintrc.js
slimee: It seems to me that yes: every async test needs that.
But we can choose to set it locally on each async test.

arnaudbesnier: I don't think so.

This test implementation won't be rejected by the initial linter rules:
```javascript
it('should call back with app', async () => {
    expect.assertions(2);

    await new Promise((resolve) => {
      const callback = (error, app) => {
        expect(error).toBeNull();
        expect(app).not.toBeNull();
        resolve();
      };
      forestExpress.init({
        ...implementation,
        opts: { callback },
      });
    });
  });
```

PR #317, opened 2019-12-18
[*] Collections - Fix included/excluded collections not removed

DrRaider:
## Pull Request checklist:

## Steps to Reproduce
Setup a project with lumber and follow the steps of the documentation https://docs.forestadmin.com/documentation/how-tos/include-exclude-models

## Expected behavior
Excluded models should disappear from the UI

## Actual behavior
Excluded models are still present but not accessible

## Corrections
The generation of the `forest` directory was updating the schema with the excluded values (or without the included ones). I added a condition on the `requireAllModels` to ignore (or only take into account) the collections the user wishes to use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixed
 - Intercom Integration - Better handling of automated messages.
+- Collections - Fix included/excluded collections on `Smart Customizations` requireAll call.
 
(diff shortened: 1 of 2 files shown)


arnaudbesnier:
Here is a simpler fix Nicolas:
https://github.com/ForestAdmin/forest-express/pull/318/files

PR #326, opened 2020-01-07
[+] Sessions - Distinguish "CORS configuration issue" and "Server down" scenarios in case of liana login error.

slimee:
This route must: 
respond on /forest/healthcheck
be always reachable (even from different domains)
is used by Liana connection to test if server is on.
It answers 200.

## Pull Request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Technical - New healthcheck route
+
 ### Fixed
 - Linter - Do not lint uncommitted files.
 - Initialization - Filter out test files when requiring models (`__tests__/*`, `*.spec.js`, `*.spec.ts`, `*.test.js` or `*.test.ts`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 ### Added
-- Technical - New healthcheck route
+- Sessions - Distinguish "CORS configuration issue" and "Server down" scenarios in case of liana login error.
 
 ### Fixed
 - Linter - Do not lint uncommitted files.
--- a/src/index.js
+++ b/src/index.js
@@ -386,4 +386,4 @@ exports.RecordRemover = require('./services/exposed/record-remover');
 exports.RecordSerializer = require('./services/exposed/record-serializer');
 exports.PermissionMiddlewareCreator = require('./middlewares/permissions');
 
-exports.PUBLIC_ROUTES = ['/', '/sessions', '/sessions-google', '/healthcheck'];
+exports.PUBLIC_ROUTES = ['/', '/healthcheck', '/sessions', '/sessions-google'];
--- a/src/index.js
+++ b/src/index.js
@@ -14,6 +14,7 @@ const AssociationsRoutes = require('./routes/associations');
 const StatRoutes = require('./routes/stats');
 const SessionRoute = require('./routes/sessions');
 const ForestRoutes = require('./routes/forest');
+const HealthCheckRoute = require('./routes/healthcheck');
 const Schemas = require('./generators/schemas');
 const SchemaSerializer = require('./serializers/schema');
 const logger = require('./services/logger');
@@ -233,6 +234,7 @@ exports.init = (Implementation) => {
         configStore.lianaOptions,
       ).perform();
     })
+    .then(() => new HealthCheckRoute(app, configStore.lianaOptions).perform())
     .then(() => new ForestRoutes(app, configStore.lianaOptions).perform())
     .then(() => app.use(pathMounted, errorHandler.catchIfAny))
     .then(() => {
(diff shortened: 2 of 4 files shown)


slimee:
[P3 - Catch CORS errors at Unlock your data and display a special error message](https://trello.com/c/m7cptH3i/3977-p3-catch-cors-errors-at-unlock-your-data-and-display-a-special-error-message)

test/routes/healthcheck.test.js
```diff
@@ -0,0 +1,24 @@
...
+        request(app)
+          .get('/forest/healthcheck')
+          .end((error, response) => {
+            expect(response.status).toStrictEqual(200);
+            expect(response.res.headers[ACCESS_CONTROL_ALLOW_ORIGIN]).toStrictEqual('*');
+            done();
+          });
+      });
+    });
+  });
```
arnaudbesnier: I think you should test '#OPTION /forest/healthcheck' to prove that the answer is always 204, whatever the domain that sends the request.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Technical - New healthcheck route
```
arnaudbesnier: I think we should describe the high level feature here.
This is not a "technical" topic as it has a real value to ForestAdmin users.

```suggestion
- Session - Distinguish "CORS configuration issue" and "Server down" scenarios in case of liana login error.
```

src/index.js
```diff
@@ -226,6 +228,7 @@ exports.init = (Implementation) => {
         configStore.lianaOptions,
       ).perform();
     })
+    .then(() => new HealthCheckRoute(app, configStore.lianaOptions).perform())
```
arnaudbesnier: Why do we mount a route/middleware twice?

src/index.js
```diff
@@ -376,4 +379,4 @@ exports.RecordRemover = require('./services/exposed/record-remover');
 exports.RecordSerializer = require('./services/exposed/record-serializer');
 exports.PermissionMiddlewareCreator = require('./middlewares/permissions');
 
-exports.PUBLIC_ROUTES = ['/', '/sessions', '/sessions-google'];
+exports.PUBLIC_ROUTES = ['/', '/sessions', '/sessions-google', '/healthcheck'];
```
arnaudbesnier: [Code style] Should we keep the routes declared in the alphabetical order?
```suggestion
exports.PUBLIC_ROUTES = ['/', '/healthcheck', '/sessions', '/sessions-google'];
```

src/index.js
slimee: bad mistake!

test/routes/healthcheck.test.js
slimee: We saw together that we are here on in a simple method so there is no need for options.

src/routes/healthcheck.js
```diff
@@ -0,0 +1,10 @@
+const cors = require('cors');
+const path = require('../services/path');
+
+module.exports = function HealthCheck(app, opts) {
+  this.perform = () => {
+    app.get(path.generate('healthcheck', opts), cors(), (request, response) => {
```
arnaudbesnier: If it is a simple request you won't need the `cors` middleware:
```suggestion
    app.get(path.generate('healthcheck', opts), (request, response) => {
```

test/routes/healthcheck.test.js
```diff
@@ -0,0 +1,24 @@
...
+  describe('#GET /forest/healthcheck', () => {
+    it('should return 200', async () => {
+      expect.assertions(2);
+      const app = createServer(envSecret, authSecret);
+      await new Promise((done) => {
+        request(app)
+          .get('/forest/healthcheck')
+          .end((error, response) => {
+            expect(response.status).toStrictEqual(200);
+            expect(response.res.headers[ACCESS_CONTROL_ALLOW_ORIGIN]).toStrictEqual('*');
```
arnaudbesnier: If it is a simple request you won't expect this header.

src/index.js
```diff
@@ -175,6 +176,7 @@ exports.init = (Implementation) => {
   }
 
   new SessionRoute(app, opts).perform();
+  new HealthCheckRoute(app, opts).perform();
```
arnaudbesnier: If possible, I would reorder them, for something more natural/logical 
```
new HealthCheckRoute(app, opts).perform();
new SessionRoute(app, opts).perform();
```
healthcheck is a very basic route without logic (basic public route).
Session routes are made for a more complex usage (public route to authenticate)

src/routes/healthcheck.js
slimee: Nope because the `origin:*` header is needed on the client-side.
Without it, the browser does not let JS access the response (neither the status code)

test/routes/healthcheck.test.js
slimee: Answered above: we need it because without it the browser doesn't let access to the response code.

PR #329, opened 2020-01-10
[*] Login - Make the login error messages brought up to the end client through toasts more accurate

adriguy:
See detailed description in the [Trello card](https://trello.com/c/xxi9pOEO/4016-logging-in-a-project-wrong-error-messages-are-shown-when-a-server-runs-on-the-correct-app-url-with-the-wrong-forestenvsecret).

## Pull Request checklist:

### DIFF
--- a/src/routes/sessions.js
+++ b/src/routes/sessions.js
@@ -1,5 +1,4 @@
 const path = require('../services/path');
-const ipWhitelist = require('../services/ip-whitelist');
 const errorMessages = require('../utils/error-messages');
 const LoginHandler = require('../services/login-handler');
 
@@ -35,8 +34,6 @@ module.exports = function Sessions(app, opts) {
       if (twoFactorRegistration && !twoFactorToken) {
         throw new Error();
       }
-
-      await ipWhitelist.retrieve(envSecret);
 
       const responseData = await new LoginHandler({
         renderingId,
--- a/src/services/authorization-finder.js
+++ b/src/services/authorization-finder.js
@@ -1,5 +1,6 @@
 const forestServerRequester = require('./forest-server-requester');
 const logger = require('./logger');
+const errorMessages = require('../utils/error-messages');
 
 function AuthorizationFinder(
   renderingId,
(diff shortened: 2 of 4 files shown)


src/routes/sessions.js
```diff
@@ -36,8 +35,6 @@ module.exports = function Sessions(app, opts) {
         throw new Error();
       }
 
-      await ipWhitelist.retrieve(envSecret);
```
adriguy: I removed this because:
- it was responsible of a generic toast message "IP Whitelist error" in case of bad env secret
- the ipWhitelist is already retrieved at the server start and at every private route, so no need to add it here.

src/services/authorization-finder.js
```diff
@@ -36,7 +37,18 @@ function AuthorizationFinder(
...
+          case errorMessages.SERVER_TRANSACTION.SECRET_AND_RENDERINGID_INCONSISTENT:
+            errorMessageToForward = error.message;
+            break;
+          case errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND:
+            errorMessageToForward = 'Cannot retrieve the project you\'re trying to unlock. '
+              + 'Please check that you\'re using the right environment secret regarding your project and environment.';
+            break;
+          default: errorMessageToForward = undefined;
+        }
+        throw new Error(errorMessageToForward);
```
adriguy: We want to throw a specific message here when a 404 or a 422 is received from the server.
In case of a 404 (SECRET_NOT_FOUND), I wrote a message more digest to be displayed in the toast for the end user.

src/services/login-handler.js
```diff
@@ -88,10 +88,6 @@ function LoginHandler({
       ).perform();
     }
 
-    if (!user) {
-      throw new Error();
-    }
-
```
adriguy: This is useless because if the user cannot be determined, it breaks before entering in that loop and the error is catched in the session.js file.

adriguy:
[Logging in a project - wrong error messages are shown when a server runs on the correct app url with the wrong FOREST_ENV_SECRET](https://trello.com/c/xxi9pOEO/4016-logging-in-a-project-wrong-error-messages-are-shown-when-a-server-runs-on-the-correct-app-url-with-the-wrong-forestenvsecret)

src/services/authorization-finder.js
larcin: I'm really not found of this, you're guessing the error code from the message, that's not robust at all. What if we handle this differently ?

adriguy: Just for the record, we talked about it and it seems that it would be a substantial refact to make it better here.
A future work session on it would be a good idea!

PR #334, opened 2020-01-14
[*] Schema -  Schemas having fields with escaped characters are now properly saved as valid JSON

jeffladiray:
Fix \r related issues & add tests

I'm not convinced by the last test of the `json.test.js` because of all the needed whitespaces, which generate ugly string. I did not found a better way to handle those (Since formatting is also a part of the `prettyPrint` function).

I've also tested `schema-file-updater.js` by generating a new lumber project with some regexp validation, which seems to produce a valid `.forestadmin-schema.json`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- Schema - Schemas having fields with escaped characters are now properly saved as valid JSON
+- Tests - Add tests for json prettyPrint
+- Tests - Add tests for schema file updater
 
 ## RELEASE 5.6.0 - 2020-01-14
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## [Unreleased]
 ### Fixed
-- Schema - Schemas having fields with escaped characters are now properly saved as valid JSON
-- Tests - Add tests for json prettyPrint
-- Tests - Add tests for schema file updater
+- Schema - Schemas having fields with escaped characters are now properly saved as valid JSON.
+- Tests - Add tests for json prettyPrint.
+- Tests - Add tests for schema file updater.
 
 ## RELEASE 5.6.0 - 2020-01-14
 ### Added
--- a/src/utils/json.js
+++ b/src/utils/json.js
@@ -44,7 +44,17 @@ const prettyPrint = (json, indentation = '') => {
   } else if (_.isNil(json)) {
     result += 'null';
   } else if (_.isString(json)) {
-    result += `"${json.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\r/g, '\\r').replace(/\n/g, '\\n')}"`;
+    // NOTICE: Escape invalid characters (see: https://www.json.org/json-en.html).
+    const escapedJsonString = json
+      .replace(/[\\]/g, '\\\\')
+      .replace(/["]/g, '\\"')
+      .replace(/[/]/g, '\\/')
+      .replace(/[\b]/g, '\\b')
+      .replace(/[\f]/g, '\\f')
+      .replace(/[\n]/g, '\\n')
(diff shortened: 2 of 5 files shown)


jeffladiray:
[P3 - Forest Schema JSON - "\" is not supported and breaks schema synchronization](https://trello.com/c/HNS7pqDd/3998-p3-forest-schema-json-is-not-supported-and-breaks-schema-synchronization)

src/utils/json.js
```diff
@@ -44,7 +44,7 @@ const prettyPrint = (json, indentation = '') => {
   } else if (_.isNil(json)) {
     result += 'null';
   } else if (_.isString(json)) {
-    result += `"${json.replace(/"/g, '\\"').replace(/\n/g, '\\n')}"`;
+    result += `"${json.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\r/g, '\\r').replace(/\n/g, '\\n')}"`;
```
rap2hpoutre: - Since the replace part is not easy to read, it could be explained with a notice
- It's hard to read on one line, we could multiline it.
- On a side note, this escapement is not complete. Some strings will still fail (see screenshot). According to this site: https://www.json.org/json-en.html, you have to escape more characters.


Here is a suggestion:

```suggestion
    // NOTICE: Escape invalid characters (see: https://www.json.org/json-en.html)
    const escapedJsonString = json
      .replace(/[\\]/g, '\\\\')
      .replace(/[\"]/g, '\\\"')
      .replace(/[\/]/g, '\\/')
      .replace(/[\b]/g, '\\b')
      .replace(/[\f]/g, '\\f')
      .replace(/[\n]/g, '\\n')
      .replace(/[\r]/g, '\\r')
      .replace(/[\t]/g, '\\t');
    result += `"${escapedJsonString}"`;
```

Screenshot:

Capture d’écran 2020-01-14 à 19 09 58

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- Schema - Schemas having fields with escaped characters are now properly saved as valid JSON
+- Tests - Add tests for json prettyPrint
+- Tests - Add tests for schema file updater
```
rap2hpoutre: We tend to add a final dot at the end of our Changelog entries.

test/services/schema-file-updater.test.js
```diff
@@ -0,0 +1,207 @@
...
+    framework: 'some-framework',
+    framework_version: 'some-framework-version',
+    orm_version: 'some-orm-version',
+  };
+  const schemaSerializer = new SchemaSerializer();
+  const { options: serializerOptions } = schemaSerializer;
+  const buildSchema = (collection, metas = meta) =>
+    new SchemaFileUpdater('test.json', collection, metas, serializerOptions).perform();
+
+  // NOTICE: Expecting fs.writeFileSync second parameter to be valid JSON
```
rap2hpoutre: Final dot is better but not that important IMO :)
```suggestion
  // NOTICE: Expecting `fs.writeFileSync` second parameter to be valid JSON.
```

test/utils/json.test.js
```diff
@@ -0,0 +1,84 @@
+const { prettyPrint } = require('../../src/utils/json');
+
+describe('utils > json', () => {
+  it('prettyPrint simple primitive types', () => {
```
rap2hpoutre: Maybe we could continue using `should` to display `it should ...`. Not sure though. You could also add another describe to cover `prettyPrint` part only.

```suggestion
  it('should pretty print simple primitive types', () => {
```

src/utils/json.js
jeffladiray: I guess I was a little bit to focused on the `\r` issue. Thanks for pointing me in that direction !
`\u` is a nightmare to handle (Since \u can represent any char, like `\u000A` is `\n`). Maybe that's why you didn't include it in the code sample.
I just had to fix lint issue related to useless char escape (like `"`), but i clearly think this solution is way better than mine!

PR #337, opened 2020-01-17
fix(security): update npm to 6.13.4 version

rap2hpoutre:
Previous version has been detected as a security issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,8 +72,9 @@
     "husky": "3.1.0",
     "jest": "24.9.0",
     "nock": "11.7.0",
+    "npm": "6.13.4",
     "onchange": "6.1.0",
-    "semantic-release": "^15.13.31",
+    "semantic-release": "15.14.0",
     "simple-git": "1.65.0",
     "sinon": "7.5.0",
     "supertest": "3.0.0"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1960,6 +1960,18 @@ bin-links@^1.1.2, bin-links@^1.1.3:
     graceful-fs "^4.1.15"
     write-file-atomic "^2.3.0"
 
+bin-links@^1.1.6:
+  version "1.1.7"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -72,8 +72,9 @@
     "husky": "3.1.0",
     "jest": "24.9.0",
     "nock": "11.7.0",
+    "npm": "6.13.4",
```
larcin: [MISC] We hadn't npm at all in the first place ? If so, why do we need it now ?

rap2hpoutre:
```
forest-express git:(chore/message-length) yarn why npm
yarn why v1.19.1
[1/4] 🤔  Why do we have the module "npm"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "npm@6.13.2"
info Reasons this module exists
   - "semantic-release#@semantic-release#npm" depends on it
   - Hoisted from "semantic-release#@semantic-release#npm#npm"
info Disk size without dependencies: "37.21MB"
info Disk size with unique dependencies: "53.14MB"
info Disk size with transitive dependencies: "69.51MB"
info Number of shared dependencies: 214
✨  Done in 1.13s.
```

larcin: 🙏

arnaudbesnier:
@rap2hpoutre Is it really necessary to add `npm` in the `devDependencies` to fix the vulnerabilities?
Shouldn't `semantic-release` reinstall do the job?
(according to this dependency rule: https://github.com/semantic-release/semantic-release/blob/master/package.json#L28)

arnaudbesnier:
I confirm it does not fix the vulnerability:
- `yarn audit` still mentions npm vulnerability.
- and Github still mentions the vulnerability
Screenshot 2020-01-22 at 10 51 16


(Good read? https://www.rockyourcode.com/fix-npm-vulnerabilities-with-yarn/)

arnaudbesnier:
I'll amend the commit on `devel` to fix this.

arnaudbesnier:
Final commit here:
https://github.com/ForestAdmin/forest-express/commit/417bd1967ebcaa2034b465a3f942fff57ecbd17e

PR #338, opened 2020-01-17
chore(technical): update PR template to conventional commits specification

rap2hpoutre:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 ## Pull Request checklist:
 
-- [ ] Write an explicit title for the Pull Request, following Conventional Commits specification
+- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
 - [ ] Test manually the implemented changes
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,6 @@
 ## Pull Request checklist:
 
-- [ ] Write an explicit title for the Pull Request
-- [ ] Write changes made in the CHANGELOG.md
+- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
```
arnaudbesnier: As the the git hook will prevent bad formatting, shouldn't we simply remove this line too?

rap2hpoutre: I think it's worth adding a link to the convention for now (for learning purpose), and maybe remove it later. Also, the developer has to remember to _name_ her commit because she will have to choose one (and review it), even if there is a bunch of commits.

Not sure though: tell me what you prefer!

arnaudbesnier: Let's keep it like that!

rap2hpoutre:
See: https://github.com/ForestAdmin/forest-express/pull/342

PR #365, opened 2020-02-27
fix: use export csv batch configuration in IdsFromRequestRetriever

rap2hpoutre:
## Description

I choose to be consistant with export csv implementation:
- batch size is 1000 (still arbitrary but now "coherent")
- order result by ID. Caveats: it only takes the first ID when there are [multiple primary keys](https://forestadmin.slack.com/archives/GAZF5Q5RV/p1582809048023700) because it's not implemented in `setOrder` methods

## Pull Request checklist:

### DIFF
--- a/src/services/ids-from-request-retriever.js
+++ b/src/services/ids-from-request-retriever.js
@@ -1,6 +1,6 @@
 const QueryDeserializer = require('../deserializers/query');
 
-const BATCH_PAGE_SIZE = 100;
+const BATCH_PAGE_SIZE = 1000;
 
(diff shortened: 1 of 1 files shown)


rap2hpoutre:
> [Suggestion] In the `records-getter.test.js`, changing the collection definition to
> 
> ```
> const collection = [
>   { id: '1', name: 'foo' },
>   { id: '3', name: 'bar' },
>   { id: '2', name: 'baz' },
> ];
> ```
> 
> would ensure that sort works at least with PK definition :)

It is ordered for technical reason only, and not in all cases: if `all_records` is `false`, there is no re-ordering . So I guess it could be misleading to un-order the list to show that the function _could_ order results.

PR #373, opened 2020-04-10
feat(integration): allow user to choose custom mapping values for intercom integration

larcin:
…ercom integration

## Pull Request checklist:

### DIFF
--- a/src/integrations/intercom/services/contact-getter.js
+++ b/src/integrations/intercom/services/contact-getter.js
@@ -1,0 +1,42 @@
+const logger = require('../../../services/logger');
+
+const EMAIL_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+function getContactQueryParameter(targetFieldValue) {
+  const targetField = targetFieldValue.match(EMAIL_REGEX) ? 'email' : 'external_id';
+
+  return {
+    query: {
+      field: targetField,
+      operator: '=',
+      value: targetFieldValue,
+    },
+  };
+}
+
+class ContactGetter {
+  static async getContact(intercomClient, Implementation, mappingValue, customerId) {
+    // NOTICE: `modelFieldName` is expected to be a sequelize/mongoose field
+    //         (ie. camelCase formatted)
+    const [modelName, modelFieldName] = mappingValue.split('.');
+    const model = Implementation.getModels()[modelName];
+    const customer = await Implementation.Intercom.getCustomer(model, customerId);
+    const targetFieldValue = customer[modelFieldName];
+
+    if (!targetFieldValue) {
+      // NOTICE: `undefined` means the field does not exist.
+      if (targetFieldValue === undefined) {
+        logger.error(`Intercom Integration Error: No field "${modelFieldName}" on model "${modelName}"`);
+      }
+      // TODO: A 404 (no intercom data for this record) should be returned here.
+      return null;
+    }
+
+    const queryParameter = getContactQueryParameter(targetFieldValue);
+
+    // TODO: Replace this by a proper call to intercom.contacts.search() when available
+    return intercomClient.post('/contacts/search', queryParameter);
+  }
+}
+
+module.exports = ContactGetter;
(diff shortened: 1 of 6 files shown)


src/integrations/intercom/services/contact-getter.js
```diff
@@ -0,0 +1,42 @@
+const logger = require('../../../services/logger');
+
+const EMAIL_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
arnaudbesnier, line 3: Looks very complex.
I know that the email format is complex, but shouldn't we simplify this and only check `@` presence (as you mentioned).
An `@` in the external id is something very unlikely to happen in my opinion, and thus, good enough.

WDYT?

src/integrations/intercom/services/contact-getter.js
```diff
@@ -0,0 +1,42 @@
...
+    query: {
+      field: targetField,
+      operator: '=',
+      value: targetFieldValue,
+    },
+  };
+}
+
+class ContactGetter {
+  static async getContact(intercomClient, Implementation, mappingValue, customerId) {
```
arnaudbesnier, line 18: Isn't the last argument the id of the record (whatever the collection name it is)?
```suggestion
  static async getContact(intercomClient, Implementation, mappingValue, recordId) {
```

src/integrations/intercom/services/contact-getter.js
```diff
@@ -0,0 +1,42 @@
...
+    const model = Implementation.getModels()[modelName];
+    const customer = await Implementation.Intercom.getCustomer(model, customerId);
+    const targetFieldValue = customer[modelFieldName];
+
+    if (!targetFieldValue) {
+      // NOTICE: `undefined` means the field does not exist.
+      if (targetFieldValue === undefined) {
+        logger.error(`Intercom Integration Error: No field "${modelFieldName}" on model "${modelName}"`);
+      }
+      // TODO: A 404 (no intercom data for this record) should be returned here.
```
arnaudbesnier, line 31: A 404 not sure.
But we could/should display an `info` log for debug purpose in the future, saying that there is no "mapping value" and so no intercom info can be retrieved in this context.

src/integrations/intercom/services/conversations-getter.js
```diff
@@ -51,56 +40,61 @@ function ConversationsGetter(Implementation, params, opts, collectionName) {
...
-        })
-        .then((conversations) => [conversations.length,
-          conversations.slice(getSkip(), getSkip() + getLimit())])
-        .spread((count, conversations) => intercom.admins.list()
+  // NOTICE: Forcing the usage of bluebird to avoid issues with the spread on route file
+  this.perform = () => P.resolve(
+    ContactGetter
+      .getContact(intercom, Implementation, mappingValue, params.recordId)
+      .then((contact) => {
+        if (!contact.body.data) {
```
arnaudbesnier, line 48: `!contact || !contact.body || !contact.body.data` ?
As you did in the "detail service" ?
Without it, I think it could raise an error here if no mapping value is found.

arnaudbesnier: I think this is what generate this ugly log
Screenshot 2020-04-10 at 18 21 56

src/integrations/intercom/services/contact-getter.js
larcin: I think there is two ways to tackle the question:
- Considering that the ids we are talking about are unlikely to contain any `@` character and thus to do a simple check as you propose
- Considering that an email as a particular structure that goes beyond the simple `@` check (see the `EMAIL_REGEX`)

But after giving a thought to those two solutions, I think I'll go your way: maybe some users will try with dummy email structure that might not match the regex (ie. `a@a.a`)

src/integrations/intercom/services/contact-getter.js
larcin: Just wanted to match the `Implementation.Intercom.getCustomer` function name but I think that function name might need to be cleaned one day (or more likely one friday afternoon)

src/integrations/intercom/services/contact-getter.js
larcin: I might remove this comment but it was kind of an hint for me if ever I want to "rework/clean" the way integrations are handled

larcin: I edited the comment nonetheless

src/integrations/intercom/services/conversations-getter.js
```diff
@@ -51,56 +40,62 @@ function ConversationsGetter(Implementation, params, opts, collectionName) {
...
-            logger.error('Cannot access to Intercom conversations: ', error);
-          }
+  // NOTICE: Forcing the usage of bluebird to avoid issues with the spread on route file
+  this.perform = () => P.resolve(
+    ContactGetter
+      .getContact(intercom, Implementation, mappingValue, params.recordId)
+      .then((contact) => {
+        if (!contact || !contact.body || !contact.body.data || !contact.body.data[0]) {
+          logger.error('Cannot access to Intercom conversations: No intercom contact matches the given key');
           return [0, []];
```
larcin, line 50: [Error handling] I don't throw errors here anymore as it ended up in an `unexpected error` instead of just a not found.

PR #375, opened 2020-04-14
feat(smart-action): allow users to protect their smart-action APIs from unauthorized usage

SteveBunlon:
## Pull Request checklist:

- [x] Test manually the implemented changes (Used CURL + UI)

I already tested this out, and it is quite hard to setup for testing. If you want, I can help you setting up everything needed to test this out 👍 

[ClickUp](https://app.clickup.com/t/4jq80f)

### DIFF
--- a/src/services/permissions-checker.js
+++ b/src/services/permissions-checker.js
@@ -5,8 +5,29 @@ const forestServerRequester = require('./forest-server-requester');
 
 let permissionsPerRendering = {};
 
-function PermissionsChecker(environmentSecret, renderingId, collectionName, permissionName) {
+function PermissionsChecker(
+  environmentSecret,
+  renderingId,
+  collectionName,
+  permissionName,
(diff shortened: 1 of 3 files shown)


src/services/permissions-checker.js
```diff
@@ -16,6 +37,9 @@ function PermissionsChecker(environmentSecret, renderingId, collectionName, perm
       return false;
     }
 
+    if (permissionName === 'smart action') {
```
arnaudbesnier, line 40: Why "smart action" and not "actions" to be consistent with the collection permission format?

arnaudbesnier: it will also be easier for developers to write "action" than "smart action" (less typo prone)

SteveBunlon: Ok I made the changes 👍

PR #393, opened 2020-05-05
fix(record serializer): harmonize the usage of the RecordSerializer

VincentMolinie:
# Description
The issue is that the `RecordSerializer` usage is different between `forest-express-sequelize` and `forest-express-mongoose`. Especially when used on SmartCollection.

# Current behaviour
```js
// On forest-express-sequelize
const serializer = new RecordSerializer({ name: 'my-smart-collection' });

// On forest-express-mongoose
const serializer = new RecordSerializer({ modelName: 'my-smart-collection' });
```

# With this PR
```js
// Whatever the liana
const serializer = new RecordSerializer({ name: 'my-smart-collection' });
```

### DIFF
--- a/src/services/exposed/record-serializer.js
+++ b/src/services/exposed/record-serializer.js
@@ -1,6 +1,18 @@
 const AbstractRecordService = require('./abstract-records-service');
 
 class RecordSerializer extends AbstractRecordService {
+  constructor(model) {
+    if (!model) {
+      throw new Error('Missing required parameters `model` passed to constructor');
+    }
+    if (!(model instanceof Object)) {
+      throw new Error('Parameter passed to RecordSerializer constructor should be an object (ex: `{ name: "myModel" }`)');
+    }
+    if (!model.modelName) {
+      model.modelName = model.name;
+    }
+    super(model);
+  }
 }
 
 module.exports = RecordSerializer;
--- a/src/services/exposed/record-serializer.js
+++ b/src/services/exposed/record-serializer.js
@@ -6,7 +6,7 @@ class RecordSerializer extends AbstractRecordService {
       throw new Error('RecordSerializer initialization error: missing first argument "model"');
     }
     if (!(model instanceof Object)) {
-      throw new Error('Parameter passed to RecordSerializer constructor should be an object (ex: `{ name: "myModel" }`)');
+      throw new Error('RecordSerializer initialization error: "model" argument should be an object (ex: `{ name: "myModel" }`)');
     }
     if (!model.modelName) {
       model.modelName = model.name;
--- a/src/services/exposed/record-serializer.js
+++ b/src/services/exposed/record-serializer.js
@@ -3,7 +3,7 @@ const AbstractRecordService = require('./abstract-records-service');
 class RecordSerializer extends AbstractRecordService {
   constructor(model) {
     if (!model) {
-      throw new Error('Missing required parameters `model` passed to constructor');
+      throw new Error('RecordSerializer initialization error: missing first argument "model"');
     }
     if (!(model instanceof Object)) {
       throw new Error('Parameter passed to RecordSerializer constructor should be an object (ex: `{ name: "myModel" }`)');


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- Record Serializer - Avoid a different usage of the serializer between `forest-express-sequelize` and `forest-express-mongoose`.
```
arnaudbesnier, line 5: This repository does not need "manual" changelog edition.
You can remove it.

arnaudbesnier:
This contribution is not a "refactor", it has an impact to the end user.
It is either a "fix" or a "feat".

arnaudbesnier:
I don't know why, but you have failing tests apparently (according to the CI)

VincentMolinie:
> I don't know why, but you have failing tests apparently (according to the CI)

I tried on local, it works ... I tried to restart the CI but it does not seems to fix the issue 🤔

src/services/exposed/record-serializer.js
```diff
@@ -1,6 +1,18 @@
 const AbstractRecordService = require('./abstract-records-service');
 
 class RecordSerializer extends AbstractRecordService {
+  constructor(model) {
+    if (!model) {
+      throw new Error('Missing required parameters `model` passed to constructor');
```
arnaudbesnier, line 6:
```suggestion
      throw new Error('Missing required first constructor argument: model');
```

rap2hpoutre:
### Breaking change

Since this service is exposed, we could consider this as a breaking change; this code **worked** (does not throw error) but **does not work anymore** (actually throws an error):

```js
const serializer = new RecordSerializer();
```

### `modelName` + `name`

With this PR, we enrich model object by adding a new property, so each model object (in sequelize and mongoose) has now two properties:

```js
{
  name: 'my-model-name',
  modelName: 'my-model-name', // Where modelName === name (not a reference anyway).
}
```

I'm not sure it's the right move, I feel like it could be considered confusing:
 - As a forest admin tool maintainer using a debugger or reading code, I don't know if I should use modelName or name.
 - This is not a reference, so there can be inconsistence.
 - We currently knows the contract, but we will forget why this actually exists in a few weeks.

Should we consider fixing (updating) forest-express-mongoose instead?

_DISCLAIMER: Maybe I missed something, I just read code in the 3 repositories so maybe I'm totally wrong. Consider this as a comment only._

src/services/exposed/record-serializer.js
```diff
@@ -1,6 +1,18 @@
...
+    }
+    if (!(model instanceof Object)) {
+      throw new Error('Parameter passed to RecordSerializer constructor should be an object (ex: `{ name: "myModel" }`)');
```
arnaudbesnier, line 9:
```suggestion
      throw new Error('RecordSerializer initialization error: "model" argument should be an object (ex: `{ name: "myModel" }`)');
```

VincentMolinie:
> ### Breaking change
> Since this service is exposed, we could consider this as a breaking change; unless I mistake, this code **worked** (does not throw error) but **does not work anymore** (actually throws an error):
> 
> ```js
> const serializer = new RecordSerializer();
> ```
> 
> ### `modelName` + `name`
> With this PR, we enrich model object by adding a new property, so each model object (in sequelize and mongoose) has now two properties:
> 
> ```js
> {
>   name: 'my-model-name',
>   modelName: 'my-model-name', // Where modelName === name (not a reference anyway).
> }
> ```
> 
> I'm not sure it's the right move, I feel like it could be considered confusing:
> 
> * As a forest admin tool maintainer using a debugger or reading code, I don't know if I should use `modelName` or `name`.
> * This is not a reference, so there can be inconsistence.
> * We currently knows the contract, but we will forget why this actually exists in a few weeks.
> 
> Should we consider fixing (updating) **forest-express-mongoose** instead?
> 
> _DISCLAIMER: Maybe I missed something, I just read code in the 3 repositories so maybe I'm totally wrong. Consider this as a comment only, food for thoughts._

- This is not a breaking change for me as initialising an empty RecordSerializer does not work ! This will crash later and makes no sense with the documentation
- I don't think there is a fix to do on `forest-express-mongoose`. It's just any time you initialize a RecordSerializer you need to pass a model or an object. As it can take a model from the constructor we need to be consistent and use the same parameter I think so `name` for sequelize and `modelName` for mongoose

PR #396, opened 2020-05-05
ci: publish releases notes on Slack

arnaudbesnier:
Uses this package:
https://github.com/juliuscc/semantic-release-slack-bot

With this Slack app:
https://api.slack.com/apps/A0132LC7EHZ/general?

With a single variable set in Travis:
Screenshot 2020-05-05 at 17 14 32

## Pull Request checklist:

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -6,5 +6,12 @@ module.exports = {
     '@semantic-release/git',
     '@semantic-release/npm',
     '@semantic-release/github',
+    [
+      'semantic-release-slack-bot',
+      {
+        notifyOnSuccess: true,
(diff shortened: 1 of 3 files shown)


arnaudbesnier:
> * (suggestion for later) I feel like publishing on #forest main channel is a bit noisy, could we consider using a dedicated channel (with CD we could have about a dozen releases a day).

I agree, we can create another webhook url at anytime to notify another channel.

> * (suggestion for later) Maybe we should be notified on release failures too (but not on the same channel), since releasing will not be manually monitored anymore.

Yes, but I don't think we can use another channel for failures with this package

PR #438, opened 2020-06-29
fix(smart-field): fix smart field name

GuillaumeCisco:
## Pull Request checklist:

Linked to clickup: https://app.clickup.com/t/5cmwz3

Link to complex implementation: https://github.com/ForestAdmin/forest-express/pull/443
Link to PR with explicit tests: https://github.com/ForestAdmin/forest-express/pull/444


GuillaumeCisco:
This is a naive implementation for fixing the issue.
It will allow sequelize magic accessors methods names to be used as smart field name, but these names will override the magic accessors methods names as seen here: https://github.com/ForestAdmin/forest-express/pull/438/files?file-filters%5B%5D=.js#diff-6986382034a8928f3bf7c5e28f0ab31cR51

Although it is not an issue in this particular case, as `setSmartFieldValue` is only called before a serialization.
I can be in the future if manipulation is needed with magic accessors just after calling `setSmartFieldValue`.
That's why I've created another way to address this issue: https://github.com/ForestAdmin/forest-express/pull/443

yarn.lock
```diff
@@ -8978,6 +8983,15 @@ send@0.17.1:
     range-parser "~1.2.1"
     statuses "~1.5.0"
 
+sequelize-mock@0.10.2:
```
arnaud-moncel, line 8986: I think you missed to run `yarn` command after your remove `sequelize-mock`

src/services/smart-fields-values-injector.js
```diff
@@ -5,7 +5,8 @@ const Schemas = require('../generators/schemas');
 
 const DEPTH_MAX_FOR_INJECTION = 0;
 
-function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0) {
+function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0,
+  requestedField = null) {
```
arnaud-moncel, line 9:
```suggestion
function SmartFieldsValuesInjector(
  record,
  modelName,
  fieldsPerModel,
  depth = 0,
  requestedField = null,
) {
```

GuillaumeCisco:
Thank you @arnaud-moncel.
I've updated the code thanks to your comments ;)

GuillaumeCisco:
Thanks a lot!
It really helps :hatching_chick:

I've looked at the code of `sequelize` for the `getDataValue` function.
It basically does: https://github.com/sequelize/sequelize/blob/master/src/model.js#L3646
```
getDataValue(key) {
    return this.dataValues[key];
  }
```

I'm looking for own properties here for being more precise, but simply using `getDataValue` could do the trick. I will test it and let you know :)

GuillaumeCisco:
I don't feel confortable replacing the `hasOwnProperty` checks to a `getDataValue` one.
Doing so would make all fields with `null`, `undefined`, `''`, `false` value to enter in the sub scope which is not what we want.

I'd like to keep it that way. More robust. If you don't mind :)

PR #461, opened 2020-07-29
chore: 🐛 fix pre-commit hook for file renames

ghusse:
Fixes the pre-commit hook that fails when there are renamed files to be committed.

## Pull Request checklist:

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Test manually the implemented changes
- [x] Review my own code (indentation, syntax, style, simplicity, readability)
- [x] Wonder if you can improve the existing code

PR #462, opened 2020-07-29
feat(error-handling): expose error-handler to be used in agents

ghusse:
This PR is based on #461, because it needed to be able to record file renames

Linked to CU-6gmxhz

The idea behind the newly exposed service, is to be able to use it in forest agents, in order to properly serialize errors when encountered.

As we don't want to write the same code multiple times, the best idea was to expose this service, and use it in lumber-generated agents.

## Pull Request checklist:


test/services/exposed/error-handler.test.js
```diff
@@ -0,0 +1,97 @@
...
+        message: 'Second message',
+      }],
+      message: 'Root message',
+    };
+
+    const next = sinon.stub();
+
+    const response = mockResponse();
+
+    const handleError = errorHandler();
```
VincentMolinie, line 37:
```suggestion
    const next = sinon.stub();
    const response = mockResponse();
    const handleError = errorHandler();
```

test/services/exposed/error-handler.test.js
```diff
@@ -0,0 +1,97 @@
...
+  it('logs the unexpected error if the error does not have a status', () => {
+    expect.assertions(2);
+
+    const error = new Error('The error');
+
+    const logger = {
+      error: sinon.stub(),
+    };
+
+    const handleError = errorHandler({ logger });
```
VincentMolinie, line 62:
```suggestion
    const error = new Error('The error');
    const logger = { error: sinon.stub() };
    const handleError = errorHandler({ logger });
```

VincentMolinie:
The title is missing a context (Always have a context in case of feature / fix). It could be `error handling`

ghusse:
> The title is missing a context (Always have a context in case of feature / fix). It could be `error handling`

Thanks, I added the context in PR's title

src/services/exposed/error-handler.js
```diff
@@ -0,0 +1,38 @@
...
+function errorHandler({ logger } = {}) {
+  /**
+   * @param {any} error
+   * @param {import('express').Request} request
+   * @param {import('express').Response} response
+   * @param {() => void} next
+   */
+  return function handleError(error, request, response, next) {
+    if (error) {
+      // NOTICE: Send the first error if any
```
guillaumedeslandes, line 16:
```suggestion
      // NOTICE: Send the first error if any.
```

PR #443, opened 2020-07-02
fix(smart-field): fix smart field name behaviour when names are reserved ones

GuillaumeCisco:
## Pull Request checklist:

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Test manually the implemented changes
- [x] Review my own code (indentation, syntax, style, simplicity, readability)
- [x] Wonder if you can improve the existing code

Linked to clickup: https://app.clickup.com/t/5cmwz3

Linked to naive implementation: https://github.com/ForestAdmin/forest-express/pull/438
Link to PR with explicit tests: https://github.com/ForestAdmin/forest-express/pull/444

PR #470, opened 2020-08-06
fix(related-data): use same reference on record for dataValues and direct attributes

GuillaumeCisco:
## Pull Request checklist:


arnaudbesnier:
Task linked: [CU-5cmwz3 P3 - Smart Field not displayed (empty cells) in the UI because of its name](https://app.clickup.com/t/5cmwz3)

src/services/smart-fields-values-injector.js
```diff
@@ -68,16 +77,22 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
...
 
           return setSmartFieldValue(record, field, modelName);
         }
         if (_.isArray(field.type)) {
-          record[field.field] = [];
+          // NOTICE: when modifying a sequelize record, make sure you pass the same reference
+          //         to record and record.dataValues
+          const emptyArray = [];
+          record[field.field] = emptyArray;
+          record.dataValues[field.field] = emptyArray;
```
GuillaumeCisco, line 95: This part of code enlightens the fix

src/services/smart-fields-values-injector.js
```diff
@@ -68,16 +77,22 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
 
   this.perform = () =>
     P.each(schema.fields, (field) => {
-      if (!record[field.field]) {
+      if (record
+        && record.dataValues
+        && !Object.prototype.hasOwnProperty.call(record.dataValues, field.field)) {
```
VincentMolinie, line 82: Why not doing this: 
```suggestion
        && !record.dataValues.hasOwnProperty(field.field)) {
```

VincentMolinie: Which I think is much cleaner

src/services/smart-fields-values-injector.js
VincentMolinie: I clearly don't think we should use `dataValues` that much. It's way too dangerous, it can breaks by just updating a minor version

src/services/smart-fields-values-injector.js
GuillaumeCisco: I wanted to do that in the first place in the original PR, but the linter does not let me :/

src/services/smart-fields-values-injector.js
GuillaumeCisco: From the moment, we manipulate and modifiy directly sequelize object, we need to reflect added attributes in `dataValues` and direct attributes. Otherwise a lot of hard bugs could happen, if one is using [`getDataValue`](https://github.com/sequelize/sequelize/blob/a2d75e6edd477b6167636d2a7e446346d666b9d9/src/model.js#L3658)
https://sequelize.readthedocs.io/en/v3/api/instance/

src/services/smart-fields-values-injector.js
ghusse: what about `field.field in record.dataValues` ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

ghusse:
I tested the regression, and it seems ok on this branch 👍

src/services/smart-fields-values-injector.js
ghusse: Ok, the explanation is here: https://eslint.org/docs/rules/no-prototype-builtins

> In ECMAScript 5.1, Object.create was added, which enables the creation of objects with a specified [[Prototype]]. Object.create(null) is a common pattern used to create objects that will be used as a Map. This can lead to errors when it is assumed that objects will have properties from Object.prototype. This rule prevents calling some Object.prototype methods directly from an object.
>
> Additionally, objects can have properties that shadow the builtins on Object.prototype, potentially causing unintended behavior or denial-of-service security vulnerabilities. For example, it would be unsafe for a webserver to parse JSON input from a client and call hasOwnProperty directly on the resulting object, because a malicious client could send a JSON value like {"hasOwnProperty": 1} and cause the server to crash.
>
> To avoid subtle bugs like this, it's better to always call these methods from Object.prototype. For example, foo.hasOwnProperty("bar") should be replaced with Object.prototype.hasOwnProperty.call(foo, "bar").

I learned something

GuillaumeCisco: Unfortunately using `!field.field in record.dataValues` won't work and will break the test :/

From the doc:
**in operator**
```
The in operator returns true if the specified property is in the specified object or its prototype chain.
...
The in operator returns true for properties in the prototype chain. (If you want to check for only non-inherited properties, use Object.prototype.hasOwnProperty() instead.)
```

**hasOwnProperty**
```
The hasOwnProperty() method returns a boolean indicating whether the object has the specified property as its own property (as opposed to inheriting it).
```

GuillaumeCisco:
> For me, it's OK, but you have to be sure about the comments made by @VincentMolinie

Thank you very much :D :surfer: 🎸

src/services/smart-fields-values-injector.js
VincentMolinie: In that case you should use the public function `setDataValue(key, value)`

GuillaumeCisco: Thanks I will update that ;)

src/services/smart-fields-values-injector.js
```diff
@@ -90,7 +105,13 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
               return null;
             }
 
-            if (!record[field.field][fieldAssociation.field]
+            if (record
+              && record[field.field]
+              && record[field.field].dataValues
```
VincentMolinie, line 110:
```suggestion
              && record[field.field].get()
```

src/services/smart-fields-values-injector.js
```diff
@@ -90,7 +105,13 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
...
+              && !Object.prototype.hasOwnProperty.call(
+                record.dataValues[field.field].dataValues,
```
VincentMolinie, line 112:
```suggestion
                record.getDataValue(field.field).get(),
```

src/services/smart-fields-values-injector.js
VincentMolinie:
```suggestion
          record.setDataValue(field.field, emptyArray);
```

src/services/smart-fields-values-injector.js
```diff
@@ -68,16 +77,22 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
 
   this.perform = () =>
     P.each(schema.fields, (field) => {
-      if (!record[field.field]) {
+      if (record
+        && record.dataValues
```
VincentMolinie, line 81:
```suggestion
        && record.get()
```

src/services/smart-fields-values-injector.js
VincentMolinie, line 82:
```suggestion
        && !Object.prototype.hasOwnProperty.call(record.get(), field.field)) {
```

test/services/smart-fields-values-injector.test.js
```diff
@@ -12,42 +13,63 @@ describe('services > smart-fields-values-injector', () => {
...
+      hasAddress: () => false,
+    };
+
+    const addressRecord = {
+      dataValues: { id: 456, user: userRecord },
+      user: userRecord,
+      hasUser: () => false,
+    };
+    const fieldsPerModel = { addresses: ['id', 'user', 'hasUser', 'smart_user'], user: ['smart', 'hasAddress'], smart_user: ['smart', 'hasAddress'] };
+    it('should inject the Smart Relationship reference', async () => {
```
VincentMolinie, line 65:
```suggestion

    it('should inject the Smart Relationship reference', async () => {
```

src/services/smart-fields-values-injector.js
GuillaumeCisco: It feel very weird to me to call `hasOwnProperty` on `.get()` instead of `.dataValues`

src/services/smart-fields-values-injector.js
```diff
@@ -68,16 +77,22 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
...
           }
 
           return setSmartFieldValue(record, field, modelName);
         }
         if (_.isArray(field.type)) {
-          record[field.field] = [];
+          // NOTICE: when modifying a sequelize record, make sure you pass the same reference
+          //         to record and record.dataValues
+          const emptyArray = [];
+          record[field.field] = emptyArray;
```
SteveBunlon, line 94: [THINKING][IGNORABLE] If I'm not mistaken, it will not be possible after this line to call record.hasOrder() (if the field name is `hasOrder`)

Is this acceptable ? What if the user is working on a smart route, use the recordGetter, and is given this record but can't work with this kind of magic accessor ?

GuillaumeCisco: Absolutely! This will override the magic accessors.
I warned about this the first time I created the solution. That's why I came with another solution using `smartValues` instead: https://github.com/ForestAdmin/forest-express/pull/443 but it add some complexity.
In those days, it was acceptable to override the magic accessor.
Here is your comment about it: https://github.com/ForestAdmin/forest-express/pull/438#pullrequestreview-445762282 ;)

Any thoughts?

SteveBunlon (approved):

PR #483, opened 2020-09-03
feat: serialize the new property isPrimaryKey

ghusse:
Linked to CU-48py0a

## Pull Request checklist:

### DIFF
--- a/src/serializers/schema.js
+++ b/src/serializers/schema.js
@@ -31,6 +31,7 @@ function SchemaSerializer() {
         'enums',
         'integration',
         'isFilterable',
+        'isPrimaryKey',
         'isReadOnly',
         'isRequired',
         'isSortable',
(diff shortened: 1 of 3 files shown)


arnaudbesnier:
Task linked: [CU-48py0a Lumber - ID fields with no default value are missing](https://app.clickup.com/t/48py0a)

DrRaider (approved):

arnaudbesnier (approved):

src/serializers/schema.js
```diff
@@ -31,6 +31,7 @@ function SchemaSerializer() {
         'enums',
         'integration',
         'isFilterable',
+        'isPrimaryKey',
```
arnaudbesnier, line 34: Actually wondering if the `isPrimaryKey` property is serialized to `true` or `false` for any field as it is for other properties (and if not, if it should)

ghusse: @arnaudbesnier I don't understand the question.

Do you want to know if there is always a value set for this property, on every field?

For what I remember, properties like `isRequired` is only set to true when a field is required, and not set at all when a field is not.

I did the same for `isPrimaryKey`, but this behavior is part of both `forest-express-sequelize` and `forest-express-mongoose`, not this package.

ghusse: Ok, I changed the code to always set the property

arnaudbesnier (approved):

PR #488, opened 2020-09-14
feat(absoluterootpath): add the option to pass the root path

### LINKED ISSUE #445: Yarn v2 breaks Schema creation
## Expected behavior

Schema is expected to be created successfully.

## Actual behavior

Errors out with `EROFS: read-only filesystem, open '/.forestadmin-schema.json'`

## Context

* Package Version: 6.3.0
* Express Version: 4.17.1
* Yarn v2

Yarn v2 has introduced a new way of storing dependencies, i.e. not in node_modules, but rather in a cache with dependencies zipped up. This breaks forest-express's getAbsolutePath utility.: https://github.com/ForestAdmin/forest-express/blob/f632e0369d03057372bec04fae13a3da83e319c8/src/utils/project-directory.js#L10

Since the dir no longer has 'node_modules', it is taking the fork `return process.cwd()`, which is returning `''` (not sure why, possibly because it's running from a zipped file?), which causes this: https://github.com/ForestAdmin/forest-express/blob/f632e0369d03057372bec04fae13a3da83e319c8/src/index.js#L35
to return an absolute path that is '/.forestadmin-schema.json'.

The temporary solution is to `yarn unplug forest-express`, but if Yarn v2 is going to become more widely used, this should probably be fixed.

julienfouilhe:
Fixes #445 

Yarn v2 sets the project dependencies in zip files in a .yarn directory. Instead of handling yarn v2 specific case, why not allow everyone to set the root directory they want to use? It would allow everyone to choose where the schema is stored.

## Pull Request checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -30,11 +30,8 @@ const ProjectDirectoryUtils = require('./utils/project-directory');
 const { is2FASaltValid } = require('./utils/token-checker');
 const { getJWTConfiguration } = require('./config/jwt');
 
-const pathProjectAbsolute = new ProjectDirectoryUtils().getAbsolutePath();
-
 const ENVIRONMENT_DEVELOPMENT = !process.env.NODE_ENV
   || ['dev', 'development'].includes(process.env.NODE_ENV);
-const SCHEMA_FILENAME = `${pathProjectAbsolute}/.forestadmin-schema.json`;
 const DISABLE_AUTO_SCHEMA_APPLY = process.env.FOREST_DISABLE_AUTO_SCHEMA_APPLY
   && JSON.parse(process.env.FOREST_DISABLE_AUTO_SCHEMA_APPLY);
 const REGEX_COOKIE_SESSION_TOKEN = /forest_session_token=([^;]*)/;
@@ -113,6 +110,8 @@ exports.init = (Implementation) => {
 
   configStore.Implementation = Implementation;
   configStore.lianaOptions = opts;
+
+  const schemaFileName = path.join(opts.absoluteRootPath || new ProjectDirectoryUtils().getAbsolutePath(), '.forestadmin-schema.json');
 
   if (opts.onlyCrudModule === true) {
     return buildSchema();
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -30,11 +30,8 @@ const ProjectDirectoryUtils = require('./utils/project-directory');
 const { is2FASaltValid } = require('./utils/token-checker');
 const { getJWTConfiguration } = require('./config/jwt');
 
-const pathProjectAbsolute = new ProjectDirectoryUtils().getAbsolutePath();
```
arnaudbesnier, line 33: In my opinion, the only function that should change is `getAbsolutePath`, as it is its ownership to return the right path according to the context of the project.

The code below can remain the same if `getAbsolutePath` return the right path in the Yarn 2 context.

src/index.js
```diff
@@ -114,6 +111,8 @@ exports.init = (Implementation) => {
   configStore.Implementation = Implementation;
   configStore.lianaOptions = opts;
 
+  const schemaFileName = path.join(opts.absoluteRootPath || new ProjectDirectoryUtils().getAbsolutePath(), '.forestadmin-schema.json');
```
arnaudbesnier, line 114: Exposing a new property `absoluteRootPath` to the agent initialisation is the last thing we want to do.
The less configuration, the easier the integration, the simpler the documentation.

Can you find a robust way to detect that Yarn 2 is used on the project to adapt what `getAbsolutePath` will return in this context?

arnaudbesnier (changes requested):

src/index.js
ArnaudRinquin: It's cool to have sane defaults but some projects setups are more complex (yarn 2 and yarn workspaces, for the most common?)

It's quite an assumption that it's always possible to have this schema file generated and loaded from a non-configurable path at the root of a project. It can be really cumbersome to play around with this arbitrary constraint.

config and models paths is configurable, that's very cool, why not allow it on this path?

julienfouilhe: Yep I agree with @ArnaudRinquin!

Also in the v7, some environment variables can be passed to the init function, but others are required in process.env because this library uses process.env directly. It's troublesome because I don't necessarily want to use the same variable name than this lib.

Making things configurable would make this lib easier to use.

I also feel like it would be better if we could pass an array of configs instead of a configDir path, but that's another topic.

ArnaudRinquin: User @merceyz [gave another great reason](https://github.com/ForestAdmin/forest-express/pull/693#discussion_r617468658) for this: with yarn 2 + `enableGlobalCache` the module can be anywhere so... this PR is a simple workaround.

julienfouilhe:
Closing in favor of #698

PR #513, opened 2020-11-05
chore: remove usage of modelsDir option

arnaud-moncel:
## Pull Request checklist:

### DIFF
--- a/src/services/config-store.js
+++ b/src/services/config-store.js
@@ -56,7 +56,7 @@ class ConfigStore {
     }
 
     if (options.modelsDir) {
-      this.logger.warn('The use of modelsDir is deprecated. Please remove the option in the Forest initialization.');
+      this.logger.warn('modelsDir is not supported anymore. Please remove this option.');
     }
   }
 }
--- a/src/services/config-store.js
+++ b/src/services/config-store.js
@@ -9,10 +9,6 @@ class ConfigStore {
     this.path = path;
   }
 
-  get modelsDir() {
-    return this.lianaOptions && this.lianaOptions.modelsDir;
-  }
-
   get configDir() {
     if (!this.lianaOptions) return null;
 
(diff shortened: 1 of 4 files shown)


src/services/config-store.js
```diff
@@ -58,6 +54,10 @@ class ConfigStore {
     if (options.onlyCrudModule) {
       this.logger.warn('The use of onlyCrudModule is deprecated. Please remove the option in Forest initialization.');
     }
+
+    if (options.modelsDir) {
+      this.logger.warn('The use of modelsDir is deprecated. Please remove the option in the Forest initialization.');
```
rap2hpoutre, line 59: (sorry for the noise: I thought about this this morning during the standup and did not want to interrupt) Is it **deprecated** or **unsupported**? It seems it's not supported anymore, so it's not deprecated. It's removed and we have a warning.

rap2hpoutre, line 59: (sorry for the noise: I thought about this this morning during the standup and did not want to interrupt) Is it **deprecated** or **unsupported**? It seems it's not supported anymore, so it's not deprecated: it's removed and we have a warning.

jeffladiray: I think you are totally right on that. In fact, we do not `deprecate` the use of it, but we definitely remove support associated to it. Rewording is needed here and #512, so ... nice catch 👍

jeffladiray (changes requested):
I tested:
- On a generated lumber project (With and without `modelsDir`)
- On our integration (Without `modelsDir`, since we're currently not using it)

And I did not see any issues.

Also, discussed with a few team member to see if they we're using `modelsDir` as an option, which does not seem to be the case. Looking at the code, I can't find any reason to keep it.

As Raphael stated, maybe you should just update the warning message to indicate that we're not supporting this option anymore instead of marking it a deprecated (IMO, it was deprecated for a long time, we're simply removing the support of it here).

Otherwise, LGTM.

jeffladiray (approved):
👍

PR #514, opened 2020-11-06
refactor: init now uses async await instead of promises

jeffladiray:
## Pull Request checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -285,8 +285,7 @@ exports.init = async (Implementation) => {
 
     if (fs.existsSync(directorySmartImplementation)) {
       await requireAllModels(directorySmartImplementation);
-    }
-    if (opts.configDir) {
+    } else if (opts.configDir) {
       logger.error('The Forest configDir option you configured does not seem to be an existing directory.');
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -94,23 +94,97 @@ exports.ensureAuthenticated = (request, response, next) => {
 
 let app = null;
 
-function buildSchema() {
+async function buildSchema() {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -94,23 +94,97 @@ exports.ensureAuthenticated = (request, response, next) => {
...
+  await Schemas.perform(
+    Implementation,
+    configStore.integrator,
+    models,
+    lianaOptions,
+  );
+  return models;
+}
+
+function generateAndSendSchema(opts) {
```
jeffladiray, line 111: This was just move to a separated function - no code changes here

src/index.js
```diff
@@ -198,26 +272,25 @@ exports.init = (Implementation) => {
...
+    } else {
+      directorySmartImplementation = `${path.resolve('.')}/forest`;
+    }
+
+    if (fs.existsSync(directorySmartImplementation)) {
+      await requireAllModels(directorySmartImplementation);
+    }
+    if (opts.configDir) {
+      logger.error('The Forest configDir option you configured does not seem to be an existing directory.');
+    }
```
arnaud-moncel, line 291: As we remove the return we should put a `else if` here ;) 
```suggestion
    if (fs.existsSync(directorySmartImplementation)) {
      await requireAllModels(directorySmartImplementation);
    } else if (opts.configDir) {
      logger.error('The Forest configDir option you configured does not seem to be an existing directory.');
    }
```

arnaud-moncel (changes requested):
Just found a little bug

arnaud-moncel (approved):
👌

PR #517, opened 2020-11-06
chore: refactor getModels function and handle exclude and include models option

arnaud-moncel:
## Pull Request checklist:

### DIFF
--- a/src/services/config-store.js
+++ b/src/services/config-store.js
@@ -48,7 +48,7 @@ class ConfigStore {
     }
 
     if (!options.connections || !options.connections.constructor.toString().match('Object')) {
-      throw new Error('The connections option seems incorectly set. Please check it is an object of named connections.');
+      throw new Error('The connections option seems incorrectly set. Please check it is an object of named connections.');
     }
 
     if (options.includedModels && !Array.isArray(options.includedModels)) {
--- a/src/services/config-store.js
+++ b/src/services/config-store.js
@@ -47,6 +47,18 @@ class ConfigStore {
       throw new Error('Your envSecret seems incorrect (64 characters required). Please check it is correctly set in your .env file.');
     }
 
+    if (!options.models || !options.models.constructor.toString().match('Object')) {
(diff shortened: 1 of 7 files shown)


src/services/config-store.js
```diff
@@ -47,6 +47,18 @@ class ConfigStore {
       throw new Error('Your envSecret seems incorrect (64 characters required). Please check it is correctly set in your .env file.');
     }
 
+    if (!options.connections || !options.connections.constructor.toString().match('Object')) {
+      throw new Error('The connections option seems incorectly set. Please check it is an object of named connections.');
```
jeffladiray, line 51:
```suggestion
      throw new Error('The connections option seems incorrectly set. Please check it is an object of named connections.');
```

jeffladiray: [Question] Do we allow `connections: {}` ? From what I can read in the test we do (And we will not trigger any warning/error), and I'm not sure that's what we want ? WDYT ?

jeffladiray (changes requested):

src/services/exposed/record-serializer.js
```diff
@@ -8,9 +8,7 @@ class RecordSerializer extends AbstractRecordService {
     if (!(model instanceof Object)) {
       throw new Error('RecordSerializer initialization error: "model" argument should be an object (ex: `{ name: "myModel" }`)');
     }
-    if (!model.modelName) {
-      model.modelName = model.name;
-    }
```
jeffladiray, line 13: You should consider re-adding theses lines since it'll break smart collection. Not sure why this fix was introduced here though, but on my side it does not work :/

jeffladiray (approved):
Pretty hard to fully test (except via automated tests, but most of your changes seems cover)

Peer reviewed and tested using new `forest-express-mongoose` & `forest-express-sequelize` with the new version, seems to work as expected. I'm pretty sure CC will complain, so let's merge this in the main PR & do a follow up to fix what need to be fixed.

NJ 🎉

PR #526, opened 2020-11-12
chore: create models manager

arnaud-moncel:
## Pull Request checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -30,6 +30,7 @@ const {
   ipWhitelist,
   apimapFieldsFormater,
   configStore,
+  modelsManager,
   fs,
 } = context.inject();
 
@@ -45,15 +46,6 @@ const TWO_FA_SECRET_SALT = process.env.FOREST_2FA_SECRET_SALT;
 
 let jwtAuthenticator;
 let app = null;
-
-function getModels() {
-  const models = configStore.Implementation.getModels();
-  _.each(models, (model, modelName) => {
-    model.modelName = modelName;
-  });
-
-  return _.values(models);
-}
 
 function loadCollections(collectionsDir) {
   try {
@@ -77,7 +69,7 @@ function loadCollections(collectionsDir) {
 
 async function buildSchema() {
   const { lianaOptions, Implementation } = configStore;
-  const models = getModels();
+  const models = modelsManager.getModelArray();
   configStore.integrator = new Integrator(lianaOptions, Implementation);
   await Schemas.perform(
     Implementation,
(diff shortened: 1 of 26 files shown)


src/index.js
```diff
@@ -77,7 +69,7 @@ function loadCollections(collectionsDir) {
 
 async function buildSchema() {
   const { lianaOptions, Implementation } = configStore;
-  const models = getModels();
+  const models = modelsManager.getModelArray();
```
jeffladiray, line 72: `getModelArray`seems weird IMO.

If I understand correctly, `modelsManager.getModels()` reads `models` in an object, and `getModelArray()`converts it to an array ?

test/services/models-manager.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+      const spy = jest.spyOn(modelsManager, '_generateModelList');
+
+      modelsManager.getModels();
+      expect(spy).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('getModelArray', () => {
+    it('should return models as array', () => {
```
jeffladiray, line 46:
```suggestion
    it('should return models as an array', () => {
```

test/services/models-manager.test.js
```diff
@@ -0,0 +1,69 @@
...
+            name: 'model2',
+          },
+        })),
+        getModelName: jest.fn((model) => model.name),
+      },
+    };
+
+    const modelsManager = new ModelsManager({ configStore });
+
+    it('should return model lists', () => {
```
jeffladiray, line 21:
```suggestion
    it('should return the model lists as an object using model name as a key', () => {
```

(Had hard time figuring this out, maybe the label of the test was ok, but a small comment on the difference between `getModels` and `getModelArray`would be nice)

src/services/models-manager.js
```diff
@@ -0,0 +1,26 @@
...
+    if (!this.models) this._generateModelList();
+    return this.models;
+  }
+
+  getModelArray() {
+    return _.values(this.getModels());
+  }
+
+  _generateModelList() {
+    const models = this.configStore.Implementation.getModels();
```
jeffladiray, line 19: If I'm right, this line is used to be compliant with the current `forest-express` version, and the goal of the following PR is to remove `Implementation` here ?

jeffladiray (commented):
Nice work 🥇 

I added a few questions and suggestions to be sure that I fully grasp the goal here (Since you split the previous PR in 2). Other than that, I still need to do a few more tests.

src/index.js
arnaud-moncel: Exactly ;) the only place where we want array is here !

src/services/models-manager.js
arnaud-moncel: Yes exactly 👍

jeffladiray (approved):
LGTM

PR #528, opened 2020-11-13
refactor: move SchemaFileUpdater to class with DI

rap2hpoutre:
## Pull Request checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,7 +21,6 @@ const Schemas = require('./generators/schemas');
 const SchemaSerializer = require('./serializers/schema');
 const Integrator = require('./integrations');
 const ApimapSender = require('./services/apimap-sender');
-const SchemaFileUpdater = require('./services/schema-file-updater');
 const ConfigStore = require('./services/config-store');
 const ProjectDirectoryUtils = require('./utils/project-directory');
 const { is2FASaltValid } = require('./utils/token-checker');
@@ -33,6 +32,7 @@ const {
   errorHandler,
   ipWhitelist,
   apimapFieldsFormater,
+  schemaFileUpdater,
 } = context.inject();
 
 const pathProjectAbsolute = new ProjectDirectoryUtils().getAbsolutePath();
@@ -155,8 +155,7 @@ function generateAndSendSchema(opts) {
       framework_version: expressVersion,
       orm_version: configStore.Implementation.getOrmVersion(),
     };
-    const content = new SchemaFileUpdater({ logger })
-      .update(SCHEMA_FILENAME, collections, meta, serializerOptions);
+    const content = schemaFileUpdater.update(SCHEMA_FILENAME, collections, meta, serializerOptions);
     collectionsSent = content.collections;
     metaSent = content.meta;
   } else {
--- a/src/index.js
+++ b/src/index.js
@@ -155,8 +155,8 @@ function generateAndSendSchema(opts) {
       framework_version: expressVersion,
       orm_version: configStore.Implementation.getOrmVersion(),
     };
-    const content = new SchemaFileUpdater(SCHEMA_FILENAME, collections, meta, serializerOptions)
-      .perform();
+    const content = new SchemaFileUpdater({ logger })
+      .update(SCHEMA_FILENAME, collections, meta, serializerOptions);
     collectionsSent = content.collections;
     metaSent = content.meta;
   } else {
--- a/src/services/schema-file-updater.js
+++ b/src/services/schema-file-updater.js
@@ -125,38 +125,48 @@ class SchemaFileUpdater {
     collection.actions = this.cleanActions(collection.actions);
   }
 
+  static formatFields(collection, serializerOptions) {
+    collection.fields = collection.fields || [];
+    collection.fields = collection.fields.map((field) => {
+      const fieldFormatted = SchemaFileUpdater
+        .formatObject(field, serializerOptions.fields.attributes);
+
+      fieldFormatted.validations = fieldFormatted.validations.map((validation) =>
+        SchemaFileUpdater.formatObject(validation, serializerOptions.validations.attributes));
+      return fieldFormatted;
+    });
+    collection.fields = _.sortBy(collection.fields, ['field', 'type']);
+  }
+
+  static formatSegments(collection, serializerOptions) {
+    collection.segments = collection.segments || [];
+    collection.segments = collection.segments.map((segment) =>
+      SchemaFileUpdater.formatObject(segment, serializerOptions.segments.attributes));
+    collection.segments = _.sortBy(collection.segments, ['name']);
+  }
+
+  static formatActions(collection, serializerOptions) {
+    collection.actions = collection.actions || [];
+    collection.actions = collection.actions.map((action) => {
+      const actionFormatted = SchemaFileUpdater
+        .formatObject(action, serializerOptions.actions.attributes);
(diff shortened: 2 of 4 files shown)


src/services/schema-file-updater.js
```diff
@@ -1,31 +1,34 @@
...
+  if (!Object.prototype.hasOwnProperty.call(object, property)) {
+    object[property] = value;
   }
+}
 
-  const cleanFields = (fields) => {
+class SchemaFileUpdater {
+  constructor({ logger }) {
+    this.logger = logger;
+  }
```
slimee, line 29: Yeah 👍 this is good.
Please note: `formatObject` and `setDefaultValueIfNecessary` should be inside the class (or inside another class if you consider it's another concern) because these functions are hidden, they are can't be tested unitary.

src/index.js
```diff
@@ -155,8 +155,8 @@ function generateAndSendSchema(opts) {
       framework_version: expressVersion,
       orm_version: configStore.Implementation.getOrmVersion(),
     };
-    const content = new SchemaFileUpdater(SCHEMA_FILENAME, collections, meta, serializerOptions)
-      .perform();
+    const content = new SchemaFileUpdater({ logger })
```
slimee, line 158: [NOT NEEDED NOW JUST TO INFORM YOU]
You can add `SchemaFileUpdater` inside the context (with `addClass` in `services-builder.js`)
And then use `inject` here to get back your instance. Having all your pieces inside the context is a good thing for reusability. 
Context building also prevent from circular dependencies, alow to test independently stuff, etc..

I set this message as NOT_NEEDED because if you follow the needs of injection, you will have to cover all the application in one PR :)

test/services/schema-file-updater.test.js
```diff
@@ -21,7 +21,7 @@ describe('services > schema-file-updater', () => {
   const schemaSerializer = new SchemaSerializer();
   const { options: serializerOptions } = schemaSerializer;
   const buildSchema = (collection, metas = meta) =>
-    new SchemaFileUpdater('test.json', collection, metas, serializerOptions).perform();
+    new SchemaFileUpdater({ logger }).update('test.json', collection, metas, serializerOptions);
 
   // NOTICE: Expecting `fs.writeFileSync` second parameter to be valid JSON.
   it('should call fs.writeFileSync with a valid JSON as data', () => {
```
slimee, line 27: [INFORMATION TO KNOW BUT WORKLOAD TO CHANGE]
This advice means a rewrite of this file, ok you can let it as is if you want, the effort to make it square doesn't bring enough, but it's important to not continue this way in future files.

This is an integrated test. When testing services, a unit test is preferable.
Unit test means testing one method only, mocking all dependencies, and to have all dependencies of the tested method injected. To test each method independently (excluding even methods of the same class)

A real integrated test of this file is when testing with the caller (method `generateAndSendSchema` in `index.js`, which is in fact the init method of forest-express).

slimee: [IMPORTANT]
You can consider create `writeFileSync` in the context.
You can do `.addInstance('writeFileSync', (args) => fs.writeFileSync(...args))`
And then injecting it in your service.

Like that in test, you will mock the write (by adding `jest.fn()`) so you can test call without a spy, and without really writing to the file which is consuming.

Then people can reuse the write-file-to-disk service your created, and tests are always the same.

slimee (commented):
This is just comments about injection

test/services/schema-file-updater.test.js
rap2hpoutre: Thank you.

I refactored to follow the `[IMPORTANT]` here: https://github.com/ForestAdmin/forest-express/pull/528/commits/2ce6ff049b8ef3cf0962a18e478e88995a518a60

I had to ignore `[INFORMATION TO KNOW BUT WORKLOAD TO CHANGE]` because it's cumbersome.

src/index.js
rap2hpoutre: `[NOT NEEDED NOW JUST TO INFORM YOU]` I considered this comment here: https://github.com/ForestAdmin/forest-express/pull/528/commits/73190b1178697af485c91d88495fbcc6b829f527

src/services/schema-file-updater.js
rap2hpoutre: Thank you. It has been fixed here: https://github.com/ForestAdmin/forest-express/pull/528/commits/cfa930dd857c5bce793b3a1ca48742dd990b3c58

rap2hpoutre:
I considered your 3 comments in three commits and added one last commit to make code climate happy: https://github.com/ForestAdmin/forest-express/pull/528/commits/9523cba122ba6a36bfdc3aab66eccbda12ecdac0

src/services/schema-file-updater.js
```diff
@@ -115,14 +120,14 @@ function SchemaFileUpdater(filename, collections, meta, serializerOptions) {
...
-    collection.segments = cleanSegments(collection.segments);
-    collection.actions = cleanActions(collection.actions);
-  };
+    collection.fields = SchemaFileUpdater.cleanFields(collection.fields);
+    collection.segments = SchemaFileUpdater.cleanSegments(collection.segments);
+    collection.actions = this.cleanActions(collection.actions);
+  }
 
-  this.perform = () => {
+  update(filename, collections, meta, serializerOptions) {
```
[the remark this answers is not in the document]
rap2hpoutre: Resolved

src/services/schema-file-updater.js
```diff
@@ -88,64 +93,67 @@ function SchemaFileUpdater(filename, collections, meta, serializerOptions) {
...
+    SchemaFileUpdater.setDefaultValueIfNecessary(collection, 'fields', []);
+    SchemaFileUpdater.setDefaultValueIfNecessary(collection, 'segments', []);
+    SchemaFileUpdater.setDefaultValueIfNecessary(collection, 'actions', []);
+
+    collection.fields = SchemaFileUpdater.cleanFields(collection.fields);
+    collection.segments = SchemaFileUpdater.cleanSegments(collection.segments);
+    collection.actions = this.cleanActions(collection.actions);
+  }
+
+  update(filename, collections, meta, serializerOptions) {
```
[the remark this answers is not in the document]
rap2hpoutre: Resolved

src/index.js
```diff
@@ -33,6 +32,7 @@ const {
   errorHandler,
   ipWhitelist,
   apimapFieldsFormater,
+  schemaFileUpdater,
 } = context.inject();
```
slimee, line 36: `inject` must be followed by
`if(!schemaFileUpdater) throw new Error('missing schemaFileUpdater')`

In order to stop now in case of dependency missing, and pointing the location.

slimee (approved):
yeah, piece of cake!

src/index.js
rap2hpoutre: Thank you! I was not aware of this rule (I guess "must" is a rule flag)
 - Does this mean it has to be done everywhere we have a `const { x, y, z } = inject()`?
 - I don't see an example of this pattern in this actual file (nor elsewhere in the repository). Does this mean `schemaFileUpdater` is special (vs `logger,pathService,errorHandler,ipWhitelist,apimapFieldsFormater` that are not tested) or is this a mistake?

slimee: I think all the repo does not follow the rule :) 
Asserting dependencies is important, yes, to early and properly fail.

PR #531, opened 2020-11-13
refactor: add error service in context

rap2hpoutre:
Required before refactoring `BaseOperatorDateParser `

## Pull Request checklist:


src/context/init.js
```diff
@@ -52,6 +53,7 @@ function initServices(context) {
   context.addInstance('errorHandler', errorHandler);
   context.addInstance('ipWhitelist', ipWhitelist);
   context.addInstance('forestServerRequester', forestServerRequester);
+  context.addInstance('errorService', errorService);
```
jeffladiray, line 56: [Suggestion] Shouldn't this also be a part of the type ?
(I totally forgot this on my first PRs, but I guess since it is in the code now, adding it would be nice)

jeffladiray (commented):
Totally unrelated to your contribution, but `services/error.js` does not look like a service to me. Maybe I'm wrong, but I would make more sense to put it as as `utils` maybe ?

It's not mandatory here, but will cost more to change in the futur I guess ?

rap2hpoutre:
@jeffladiray Thank you for your suggestions, I considered both.
 - error service is not a service anymore (moved to utils)
 - added type hint

🙏 

_PS: Code Climate is grumpy, but I guess we could ignore this refactor advice, it's just because I move the file._

jeffladiray (approved):
LGTM

PR #535, opened 2020-11-16
feat(smart actions): add hooks in schema file

rap2hpoutre:
This PR covers these points from this [EPIC](https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/eJGoNND3JQ#) ([clickup task](https://app.clickup.com/t/a6yc0v)):

- [x] Add `hooks` for smart actions in `.forestadmin-shema` as an object with two properties:
  - [x] `load` (`boolean`, default `false`). Considered true when `hooks: { load: something }` is declared and is not falsy.
  - [x] `change` (`array`, default `[]`). Contains each keys when `hooks: { change }` is present and `change` is an object.
- [x] Serializer (`src/serializer/schema.js`) must handle hooks property **forest-express** (and thus it has to be sent to **forestadmin-server**)

ℹ️  This PR should be rebased to master after this one is merged (refactoring to ease tests): https://github.com/ForestAdmin/forest-express/pull/528

## Pull Request checklist:

### DIFF
--- a/test/services/schema-file-updater.test.js
+++ b/test/services/schema-file-updater.test.js
@@ -165,12 +165,14 @@ describe('services > schema-file-updater', () => {
   });
 
   it('should format action hooks', () => {
-    expect.assertions(4);
+    expect.assertions(6);
 
     const INVALID_HOOKS = 'Action With Invalid Hooks';
     const INVALID_LOAD_HOOK = 'Action With Invalid Load Hook';
     const INVALID_CHANGE_HOOK = 'Action With Invalid Change Hook';
     const VALID_HOOKS = 'Action With Valid Hooks';
+    const VALID_ONLY_LOAD_HOOK = 'Action With Valid Only a Load Hook';
+    const VALID_ONLY_CHANGE_HOOK = 'Action With Valid Only a Change Hook';
 
     const schema = buildSchema([{
       name: 'collectionName',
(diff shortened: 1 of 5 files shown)


GuillaumeCisco (commented):
Thanks, this is great! ✨ 
By reading the code, it seems ok to me 👍 

Could we just add some tests with sometimes only `load` hook and sometimes only `change` hook? To be sure to know how it will behave 🚗

rap2hpoutre:
> Could we just add some tests with sometimes only load hook and sometimes only change hook? To be sure to know how it will behave 🚗

👍  Sure!

See: https://github.com/ForestAdmin/forest-express/pull/535/commits/01720e866abc7c251571a7187dadd70d1ba37ded

test/services/schema-file-updater.test.js
```diff
@@ -163,6 +164,85 @@ describe('services > schema-file-updater', () => {
...
+      .toMatchObject({ hooks: { load: false, change: ['foo'] } });
+
+    expect(findAction(INVALID_CHANGE_HOOK))
+      .toMatchObject({ hooks: { load: true, change: [] } });
+
+    expect(findAction(VALID_HOOKS))
+      .toMatchObject({ hooks: { load: true, change: ['foo', 'bar'] } });
+
+    expect(findAction(VALID_ONLY_LOAD_HOOK))
+      .toMatchObject({ hooks: { load: true, change: [] } });
```
GuillaumeCisco, line 240: shouldn't be `false` in this case?

GuillaumeCisco (commented):
Just a quick question to be sure what we should do.

Everything is ok to me

test/services/schema-file-updater.test.js
rap2hpoutre: It seems the test is correct:

 - `VALID_ONLY_LOAD_HOOK ` is an _"Action With Only a Load Hook"_, so it aims to have a load hook but no change hook.
 - Then it is initialized with: `hooks: { load: () => { } },`
 - The result is what we expect:
   - `load` it `true` (because it has a function)
   - `change` is `[]` which is the default value of `change`

The objective was to have a `load` hook set and the default value for `change` hook, everything seems fine.

👉 I just improved the description of the actions

GuillaumeCisco (approved):
Great 🚀 
LGTM ✨

PR #539, opened 2020-11-16
refactor: add schemas generator in init

rap2hpoutre:
I need `schemas` generator as inject to refactor `src/routes/actions.js` to develop this feature: https://app.clickup.com/t/a6yc11. PR waiting for this: https://github.com/ForestAdmin/forest-express/pull/538

## Pull Request checklist:

### DIFF
--- a/test/generators/schemas.test.js
+++ b/test/generators/schemas.test.js
@@ -1,0 +1,21 @@
+const schemasGenerator = require('../../src/generators/schemas');
+
+describe('generators > schemas', () => {
+  it('should build schemas', async () => {
+    expect.assertions(2);
+    const implementation = {
+      SchemaAdapter: async () => ({}),
+      getModelName: jest.fn(() => 'model name'),
+    };
+    const integrator = {
+      defineFields: jest.fn(),
+      defineSegments: jest.fn(),
+    };
+    const models = [{}, {}];
+    const opts = {};
+    await schemasGenerator.perform(implementation, integrator, models, opts);
+
+    expect(integrator.defineFields).toHaveBeenCalledTimes(models.length);
+    expect(integrator.defineSegments).toHaveBeenCalledTimes(models.length);
+  });
+});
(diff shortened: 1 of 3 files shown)


test/generators/schemas.test.js
```diff
@@ -0,0 +1,96 @@
+const uuidV1 = require('uuid/v1');
+const schemasGenerator = require('../../src/generators/schemas');
+
+
+describe('generators > schemas', () => {
```
SteveBunlon, line 5: [HUGE] This is really great you tested it 👍 

[IGNORABLE] However, I tried to deep dive in the test to understand more precisely what that generator takes as inputs, and what it returns as output, I couldn't understand easily TBH. 

Using fake models like `[{}, {}]` does not give a clear vision about the data structures implied here. Knowing that you already brainstormed on this service, I'm wondering if putting some more of your knowledge here by specifying more precisely the input data structures would not profit for the next developer who'll work on this.

SteveBunlon (approved):
Nice convention application and nice testing 👍

test/generators/schemas.test.js
rap2hpoutre: `models` are not used in `schemasGenerator`. They are just passed to `implementation.SchemaAdapter()` that returns a `schema`. This `schema` is the actual object that is used, and it contains values in tests: https://github.com/ForestAdmin/forest-express/pull/539/files#diff-28e774ccc2b4838827f6fe01fb3d68ff0a75e54f836ae9a1389fb927668c5ef4R35-R38

So we don't want to test `SchemaAdapter` which is another module. It is mocked and return the `schema` that we want to actually test. 

TL;DR: model is useless in `schemasGenerator`. We don't need it to test. Still, we need to mock the result of `implementation.SchemaAdapter()`

rap2hpoutre: The thing is that these test checks all cases in the actual file refactored:

> The test coverage on the diff in this pull request is 100.0%

We have to refer to the test case themselves to check what was tested:

>   generators > schemas
    ✓ should build schemas (5ms)
    ✓ should merge fields, actions, segments and searchFields when same model is called twice (1ms)
    ✓ should not mix models (1ms)

PR #542, opened 2020-11-17
chore: fix context creation

arnaud-moncel:
## Pull Request checklist:

EDIT: Fixes an issue around context on `master`, reproductible by linking `forest-express` in forest-express-sequelize` then in a lumber project.


test/services/permissions.test.js
```diff
@@ -1,6 +1,9 @@
 const nock = require('nock');
 const PermissionsChecker = require('../../src/services/permissions-checker');
 const context = require('../../src/context/');
+const initContext = require('../../src/context/init');
+
+context.init(initContext);
```
guillaumedeslandes, line 6: This init (and the two below at the top level describe) should be done in a before hook.

arnaud-moncel: I'm totally agree with you, but the hooks is not allowed in `forest-express`

arnaud-moncel: I can move this in the describe function instead ?

guillaumedeslandes (approved):

PR #546, opened 2020-11-20
feat(smart actions): endpoint that handle forms' load hooks

rap2hpoutre:
## Developer' agents have an endpoint that handle smart actions forms' `load` hooks

- [x] The routes verb is `POST`
- [x] The routes path is: `actions/smart-action-endpoint/hooks/load` (or the smart action endpoint)
- [x] The routes handles parameters from request
  - [x] It should transform a `recordId` to a record.
  - [x] It should get `fields` from request or get it from schema when missing (we could discuss about it later).
- [x] The routes sends `fields` and `record` to the callback (the function defined by the user).
- [x] The route has to consider the return value from the callback then trigger error in UI with 500 code and console when something went wrong, e.g:
  - [x] Error has been thrown in the function
  - [x] Nothing is returned
  - [x] `load` is not a function
  - [x] a field is deleted or added (new key or missing key in `fields`)
  - [x] Add a test for each error case
- [x] Add a test for valid case

Almost all lines of code are covered (90.3%, not sure why it's not 100%).

I suggest reviewing code by opening `src/routes/actions.js` rather than reading the diff.

## Pull Request checklist:

- [ ] Test manually the implemented changes _(not that simple, I will do this during review)_

## Tests

```
routes > actions
  ✓ should not create a route when no actions is present (3ms)
  ✓ should not create a route when actions.values or actions.hooks.* are missing present (1ms)
  when actions.values is present
    ✓ should create a route (1ms)
    ✓ should create a valid route callback (1ms)
    ✓ should handle async values function (1ms)
  when action.hooks is present
    when action.hooks.load is present
      ✓ should create a route
      when calling the route controller
        when action.hooks.load is invalid
          ✓ should fail with message when action.hooks.load is not a function (2ms)
          ✓ should fail with message when action.hooks.load does not return an object (2ms)
          ✓ should fail with message when action.hooks.load returned fields are not consistent
        when action.hooks.load is valid
          ✓ should respond success with the updated fields
```


src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+      try {
+        if (typeof action.hooks.load !== 'function') throw new Error('load must be a function');
+
+        const result = await action.hooks.load({ record, fields });
+
+        if (!(result && typeof result === 'object')) {
+          throw new Error('load hook must return an object');
+        } else if (!Actions.areFieldsConsistant(result, fields)) {
+          throw new Error('fields must be unchanged (no addition nor deletion allowed)');
+        }
```
GuillaumeCisco, line 80: We also need to check no extra properties on fields have been added or some have been removed.

test/routes/actions.test.js
```diff
@@ -20,6 +20,40 @@ function initContext(schema) {
...
+        hooks: { load },
+        fields: [{ field: 'invoice number', type: 'String' }],
+      }],
+    },
+  };
+  const {
+    actions, model, app, logger,
+  } = initContext(schemas).inject();
+
+  const request = { body: { data: { attributes: { recordsId: [1] } } } };
```
GuillaumeCisco, line 37: As said previously, I think it will be:
`const request = { body:  { recordsId: [1] } };`

test/routes/actions.test.js
```diff
@@ -36,7 +70,7 @@ describe('routes > actions', () => {
     expect(app.post).not.toHaveBeenCalled();
   });
 
-  it('should not create a route when no actions.values is present', async () => {
+  it('should not create a route when actions.values or actions.hooks.* are missing present', async () => {
```
GuillaumeCisco, line 73: is `present` needed here?

test/routes/actions.test.js
```diff
@@ -125,4 +159,102 @@ describe('routes > actions', () => {
...
+            expect(implementation.ResourceGetter)
+              .toHaveBeenNthCalledWith(1, model, { recordId: 1 });
+            expect(response.status).toHaveBeenNthCalledWith(1, 200);
+            expect(load).toHaveBeenNthCalledWith(1, {
+              fields: {
+                'invoice number': { field: 'invoice number', type: 'String', value: null },
+              },
+              record: { id: 1, name: 'Jane' },
+            });
+            expect(send).toHaveBeenNthCalledWith(1, { fields: newFields });
```
GuillaumeCisco, line 254: It should be:
```
expect(send).toHaveBeenNthCalledWith(1, { fields: [{ field: 'invoice number', type: 'String', value: 'hello from load' }] });
```

Otherwise, we will lose the position in the frontend.

GuillaumeCisco (commented):
Thank you for your work! 🔥 
Great thing we can go further in this topic.

I added some possible improvements and some questions.

src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+  /**
+   * Compare two sets of fields keys, returns true if both have the same keys.
+   * An error has to be thrown in getHookLoadController() when a field is deleted or added
+   * (new key or missing key in fields).
+   *
+   * @param {Object} fields The fields we want to compare
+   * @param {Object} previousFields The fields we want to compare
+   * @see getHookLoadController() for usage
+   */
+  static areFieldsConsistant(fields, previousFields) {
```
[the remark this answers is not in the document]
rap2hpoutre: 👍 I don't like it to be static either TBH.

The linter complains when there is no `this` and that it's not static (I can't commit). I can either:
 - keep it static (as the linter suggest)
 - add an exception with a `/* eslint-ignore .... */`

What do you suggest?

src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+  getValuesController(action) {
     return (request, response) => {
-      let values = action.values ? action.values(request.body.data.attributes.values) : {};
+      const successResponse = (object) => response.status(200).send(object);
+      const values = action.values ? action.values(request.body.data.attributes.values) : {};
 
-      if (!(values.then && typeof func === 'function')) {
-        return response.status(200).send(values);
-      }
+      if (!(values.then && typeof func === 'function')) return successResponse(values);
```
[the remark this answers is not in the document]
rap2hpoutre: You are right, it makes no sense! Nice catch, it has been done **by me** in a previous commit :/. It's a bad copypasta.

I have to fix it here, thanks!

src/routes/actions.js
rap2hpoutre: `areFieldsConsistant` validate this case.

src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+        } else if (!Actions.areFieldsConsistant(result, fields)) {
+          throw new Error('fields must be unchanged (no addition nor deletion allowed)');
+        }
+
+        return response.status(200).send({ fields: result });
+      } catch ({ message }) {
+        this.logger.error('Error in smart action load hook: ', message);
+        return response.status(500).send({ message });
+      }
+    };
```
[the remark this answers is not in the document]
rap2hpoutre: Haha I have done this because code climate said there are too much return 🙃

src/routes/actions.js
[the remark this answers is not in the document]
GuillaumeCisco: Can we put it outside of the class?

src/routes/actions.js
GuillaumeCisco: Does `areFieldsConsistant` check if there is no addition/deletion of fields AND no addition/deletion of fields properties?
This is linked to: https://github.com/ForestAdmin/forestadmin/pull/2688/files#diff-0e15c5d2b7cb907c2a110c883d828d030e3cb31c75f66a44858a11a89d13d1ddR49

src/routes/actions.js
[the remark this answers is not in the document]
GuillaumeCisco: Does refactoring it this way and splitting code with new function help?

test/routes/actions.test.js
rap2hpoutre: Fixed.

test/routes/actions.test.js
rap2hpoutre: Fixed

test/routes/actions.test.js
rap2hpoutre: Fixed

src/routes/actions.js
rap2hpoutre: > no addition/deletion of fields properties?

It is not that simple due to various reasons. let's talk about it when you want, then we will make a decision.

src/routes/actions.js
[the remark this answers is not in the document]
rap2hpoutre: Fixed via refactoring

src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+   * Generate a callback for express that handles the `load` hook.
+   *
+   * @param {*} action The smart action
+   * @param {*} model The model of the smart action
+   * @param {*} Implementation Gives access to current Implementation (mongoose or sequelize)
+   * @returns {Function} A route callback for express
+   */
+  getHookLoadController(action, model, Implementation) {
+    return async (request, response) => {
+      const recordId = request.body.data.attributes.recordsId[0];
```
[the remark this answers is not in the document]
rap2hpoutre: FixeD.

src/routes/actions.js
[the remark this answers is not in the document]
rap2hpoutre: Fixed

src/routes/actions.js
```diff
@@ -10,39 +13,130 @@ class Actions {
...
+   *
+   * @param {Object} fields The fields we want to compare
+   * @param {Object} previousFields The fields we want to compare
+   * @see getHookLoadController() for usage
+   */
+  static areFieldsConsistant(fields, previousFields) {
+    const containsAll = (left, right) =>
+      Object.keys(left).every((key) => Object.keys(right).includes(key));
+    return containsAll(fields, previousFields) && containsAll(previousFields, fields);
+  }
```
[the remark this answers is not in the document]
rap2hpoutre: Fixed.

src/routes/actions.js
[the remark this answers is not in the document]
rap2hpoutre: Yes. Fixed!

src/routes/actions.js
rap2hpoutre: Resolved

src/services/hook-load.js
```diff
@@ -0,0 +1,54 @@
...
+    const updatedFields = fields.map((field) => result[field.field]);
+
+    // Check if fields properties have changed.
+    if (this.haveFieldsPropertiesChanged(fieldsWithValue, updatedFields)) {
+      throw new Error('fields properties must be unchanged (no addition nor deletion allowed)');
+    }
+
+    // Apply result on fields (transform the object back to an array), preserve order.
+    return updatedFields;
+  }
```
GuillaumeCisco, line 39: We can simplify this method with :
```javascript
if (typeof hook !== 'function') throw new Error('load must be a function');

    // NOTICE: ease fields data structure for user
    const fieldsForUser = fields.reduce((previous, current) => ({
      ...previous,
      [current.field]: {
        ...current, // FIXME: do we remove field? Can user use it?
        value: null,
      },
    }), {});

    // Call the user-defined load hook.
    const result = await hook({ record, fields: fieldsForUser });
    if (typeof result !== 'object') {
      throw new Error('load hook must return an object');
    } else if (!this.isSameDataStructure(fieldsForUser, result)) {
      throw new Error('fields must be unchanged (no addition nor deletion allowed)');
    }

    return fields.map(field => result[field.field]);
```

`isSameDataStructure` is defined in the next comment

src/routes/actions.js
```diff
@@ -1,48 +1,121 @@
...
+          record,
+        );
+
+        return response.status(200).send({ fields: updatedFields });
+      } catch ({ message }) {
+        this.logger.error('Error in smart action load hook: ', message);
+        return response.status(500).send({ message });
+      }
+    };
+  }
```
GuillaumeCisco, line 67:
```suggestion
      let status = 200;
      let data = null;
      try {
        const updatedFields = await this.hookLoad.getResponse(
          action.hooks.load,
          action.fields,
          record,
        );

        data = { fields: updatedFields };
      } catch ({ message }) {
        this.logger.error('Error in smart action load hook: ', message);
        data = { message };
        status = 500;
      }
      
      return response.status(status).send(data);
    };
  }
```

GuillaumeCisco (commented):
With this refacto, code is far more easier to read.
I've added some improvements.
Especially in term of complexity.

I'm available for talking about it :)

src/utils/objects-have-same-keys.js
```diff
@@ -0,0 +1,16 @@
...
+   * @param  {...any} objects The list of fields object we want to compare
+   * @see HookLoad.getResponse() for usage
+ */
+function objectsHaveSameKeys(...objects) {
+  const allKeys = objects.reduce((keys, object) => keys.concat(Object.keys(object)), []);
+  const union = new Set(allKeys);
+  return objects.every((object) => union.size === Object.keys(object).length);
+}
+
+module.exports = objectsHaveSameKeys;
```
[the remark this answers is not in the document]
GuillaumeCisco: We can even go further and simplify with:
```javascript
const isSameDataStructure = (fields, previousFields, deep = 0, step = 0) => {
    if (!fields
        || !previousFields
        || !(typeof fields === 'object' && typeof previousFields === 'object')) {
        return false;
    }

    const fieldsKeys = Object.keys(fields); // O(n)
    const previousFieldsKeys = Object.keys(previousFields);  // O(m)
    if (fieldsKeys.length !== previousFieldsKeys.length) {
        return false;
    }

    // NOTICE: use Set because `has` method has a time complexity of `O(1)`.
    const fieldsKeysSet = new Set(fieldsKeys);  // O(n)
    for (const key of previousFieldsKeys) {  // O(m)
        // NOTICE: if key does not exist in other object or by its children.
        if (!fieldsKeysSet.has(key)
            || (step + 1 <= deep   // recursive, limit to deep
                && !isSameDataStructure(fields[key], previousFields[key], deep, step + 1))) {
            return false;
        }
    }

    return true;
}
```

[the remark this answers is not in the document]
rap2hpoutre: Let's go with this.

src/utils/is-same-data-structure.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Object} other the second object to compare
+ * @param {Number} deep comparison level (deepness)
+ * @param {Number} step current step (internal var)
+ * @returns {Boolean}
+ */
+function isSameDataStructure(object, other, deep = 0, step = 0) {
+  if (!object
+    || !other
+    || !(typeof object === 'object' && typeof other === 'object')) {
+    return false;
```
GuillaumeCisco, line 15:
```js
expect(isSameDataStructure(
        null,
        { a: 'a' },
      )).toBe(true);
```
would test it I think

GuillaumeCisco (approved):
LGTM 🚀 

I will check if we can get rid of the codeclimate warning.
seems overkill to me.

src/utils/is-same-data-structure.js
rap2hpoutre: Just added:

```
expect(isSameDataStructure(null, { a: 'a' })).toBe(false);
```

src/services/hook-load.js
rap2hpoutre: DOne.

src/utils/is-same-data-structure.js
```diff
@@ -0,0 +1,39 @@
...
+ * Check if two objects have the same structure.
+ * See: https://github.com/ForestAdmin/forest-express/pull/546#discussion_r530205698
+ *
+ * @param {Object} object the first object to compare
+ * @param {Object} other the second object to compare
+ * @param {Number} deep comparison level (deepness)
+ * @param {Number} step current step (internal var)
+ * @returns {Boolean}
+ */
+function isSameDataStructure(object, other, deep = 0, step = 0) {
```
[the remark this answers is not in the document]
rap2hpoutre: Ignored.

GuillaumeCisco (approved):
LGTM ✨

PR #550, opened 2020-11-23
ci: enable beta release via semantic-release

arnaud-moncel:
## Pull Request checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
       after_script:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
     - stage: Build & Publish
-      if: branch IN ("master", "beta", "alpha") AND type != pull_request
+      if: branch IN ("master", "beta") AND type != pull_request
       script: skip
       deploy:
         provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,9 @@ jobs:
       after_script:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
     - stage: Build & Publish
-      if: branch = master AND type != pull_request
+      if: branch IN ("master", "beta", "alpha") AND type != pull_request
       script: skip
       deploy:
         provider: script
         skip_cleanup: true
-        on:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,11 +21,9 @@ jobs:
...
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
     - stage: Build & Publish
-      if: branch = master AND type != pull_request
+      if: branch IN ("master", "beta", "alpha") AND type != pull_request
       script: skip
       deploy:
         provider: script
         skip_cleanup: true
-        on:
-          branch: master
```
jeffladiray, line 30: Not sure why that would be mandatory here, since this step should only be run in the `master` context, that's why we removed it

.travis.yml
```diff
@@ -21,11 +21,9 @@ jobs:
       after_script:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
     - stage: Build & Publish
-      if: branch = master AND type != pull_request
+      if: branch IN ("master", "beta", "alpha") AND type != pull_request
```
jeffladiray, line 24: `beta` channel is configured by default on `semantic-release`, so running `npx semantic-release` should be enough (See: https://semantic-release.gitbook.io/semantic-release/usage/configuration#branches)

arnaudbesnier: ok for "beta".
Is "alpha" really necessary here? Do we plan to create an "alpha" branch?

.travis.yml
arnaudbesnier: Maybe it was an unnecessary defensive condition.

.travis.yml
jeffladiray: It is not necessary, but we've added it since it has the same pre-release behavior. We will remove it 👍

jeffladiray (approved):

rap2hpoutre (commented):
Let's try this! Maybe it would generate issues, however, `forest-express` is a good candidate for testing beta releases without breaking our public/internal official releases.

rap2hpoutre (approved):

PR #552, opened 2020-11-23
ci: change semantic-release to use conventional commit preset

arnaud-moncel:
This PR change the preset used by semantic release to handle properly our commit nomenclature.
An example of unhandled commit before:
```bash
feat!: my awesome feature
```

Now it is more consistent with our commit lint message who use conventional commit.
Here is the result of generated notes for change-log
image

## Pull Request checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "@semantic-release/git": "9.0.0",
     "@types/jest": "26.0.13",
     "babel-eslint": "10.0.3",
+    "conventional-changelog-conventionalcommits": "4.5.0",
     "eslint": "6.8.0",
     "eslint-config-airbnb-base": "14.0.0",
     "eslint-plugin-import": "2.18.2",
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -1,7 +1,7 @@
 module.exports = {
   plugins: [
-    '@semantic-release/commit-analyzer',
-    '@semantic-release/release-notes-generator',
+    ['@semantic-release/commit-analyzer', { preset: "conventionalcommits" }],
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
     "@semantic-release/git": "9.0.0",
     "@types/jest": "26.0.13",
     "babel-eslint": "10.0.3",
+    "conventional-changelog-conventionalcommits": "4.5.0",
```
arnaudbesnier, line 71: I thought the only way to define a breaking change was this one:
https://github.com/semantic-release/semantic-release#commit-message-format

with a description of the breaking change in the commit message body.

Are we sure, we want to enable breaking changes to be committed without explanation (ie using `!`)?

arnaudbesnier (commented):
`ci` is a valid commit prefix?

Not convinced by the idea of committing breaking changes without explaining why it breaks.
But I understand that we are currently in the middle of both world if the `!` was allowed in commit messages prefixes.

package.json
jeffladiray: I think you are right on the fact that a BREAKING CHANGE should come with a description, making this contribution un-helpful. 

Let's update our commit on the `beta` branch to add a BREAKING CHANGE section in the body and re-check that the pre-release happen this time.

jeffladiray:
> ci is a valid commit prefix?

`ci` as a type is allowed by `commitlint` using the angular preset, but not by the conventionnal commit standard (See https://github.com/conventional-changelog/commitlint#what-is-commitlint vs https://www.conventionalcommits.org/en/v1.0.0/).

PR #563, opened 2020-11-26
feat(role-permissions): request the new permissions format

adriguy:
## Pull Request checklist:


src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
+  }
+
+  static transformPermissionsFromOldToNewFormat(permissions) {
+    Object.keys(permissions).forEach((modelName) => {
+      const { collection } = permissions[modelName];
+      permissions[modelName].collection = {
+        browseEnabled: collection.list || collection.searchToEdit,
+        readEnabled: collection.show,
+        editEnabled: collection.create,
+        addEnabled: collection.update,
```
VincentMolinie, line 57: Ahahah you switch the value 😂

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
+      && PermissionsChecker.permissions.renderings[this.renderingId].data[collectionName]) {
+      return PermissionsChecker.permissions.renderings[this.renderingId].data[collectionName].scope;
+    }
+    return null;
+  }
+
+  static transformPermissionsFromOldToNewFormat(permissions) {
+    Object.keys(permissions).forEach((modelName) => {
+      const { collection } = permissions[modelName];
+      permissions[modelName].collection = {
```
VincentMolinie, line 53: define a new object instead of reusing the same object.
For 2 reasons:
- Your function is modifying a parameter which is really bad. We don't want that.
- Your permissions will actually contain the new AND the old format 😅

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
+        browseEnabled: collection.list || collection.searchToEdit,
+        readEnabled: collection.show,
+        editEnabled: collection.create,
+        addEnabled: collection.update,
+        deleteEnabled: collection.delete,
+        exportEnabled: collection.export,
+      };
+
+      if (permissions[modelName].actions) {
```
VincentMolinie, line 62: You are using `permissions[modelName]` like a thousand times 😅  you should define a variable 😉

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
+
+  static getLastRetrieveTime(renderingId, permissionName) {
+    // In the case of rolesACL format and browseEnabled permission, the last retrieve to be taken
+    // into account is the one stored by rendering (because of the scope information).
+    if (PermissionsChecker.isrolesACLActivated && permissionName !== 'browseEnabled') {
+      return PermissionsChecker.permissions.collections.lastRetrieve;
+    }
+    return PermissionsChecker.permissions.renderings[renderingId]
+      ? PermissionsChecker.permissions.renderings[renderingId].lastRetrieve
+      : null;
```
VincentMolinie, line 113: This is due to the fact that you use 2 different formats when storing your data. You should not change this function

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
   static resetExpiration(renderingId) {
-    const permissions = PermissionsChecker.getPermissions(renderingId);
-    if (permissions) {
-      permissions.lastRetrieve = null;
+    if (PermissionsChecker.isrolesACLActivated && PermissionsChecker.permissions.collection) {
+      PermissionsChecker.permissions.collection.lastRetrieve = null;
+    }
+
+    if (PermissionsChecker.permissions.renderings[renderingId]) {
+      PermissionsChecker.permissions.renderings[renderingId].lastRetrieve = null;
```
VincentMolinie, line 122: Nor this one !

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
     }
   }
 
+  static _isPermissionAllowed(permissionValue, userId) {
+    return Array.isArray(permissionValue)
+      ? permissionValue.includes(parseInt(userId, 10))
```
VincentMolinie, line 128:
```suggestion
      ? permissionValue.includes(Number.parseInt(userId, 10))
```

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
...
-    const computedConditionFilters = _.clone(collectionListScope.filter);
+  static _computeConditionFiltersFromScope(userId, scope) {
+    const computedConditionFilters = _.clone(scope.filter);
     computedConditionFilters.conditions.forEach((condition) => {
       if (condition.value
         && `${condition.value}`.startsWith('$')
-        && collectionListScope.dynamicScopesValues.users[userId]) {
-        condition.value = collectionListScope
+        && scope.dynamicScopesValues.users[userId]) {
+        condition.value = scope
```
VincentMolinie, line 156: This looks like a refacto and should not be part of this PR

src/services/permissions-checker.js
```diff
@@ -101,89 +187,103 @@ class PermissionsChecker {
...
-        && expectedConditionFilters.conditions.length === 1;
+  static async _isCollectionBrowseAllowed(collectionPermissions, permissionInfos, scope) {
+    const { browseEnabled } = collectionPermissions.collection;
+    const { userId } = permissionInfos;
 
-      const isSameScope = !!scopeFound
-        && scopeFound.aggregator === expectedConditionFilters.aggregator
-        && !!scopeFound.conditions
-        && scopeFound.conditions.length === expectedConditionFilters.conditions.length;
+    if ((Array.isArray(browseEnabled) && !browseEnabled.includes(parseInt(userId, 10)))
```
VincentMolinie, line 229: You've created a function that does exactly that 😅 `_isPermissionAllowed` 🤔

test/services/permissions.test.js
```diff
@@ -9,154 +9,164 @@ describe('services > permissions', () => {
...
+                  },
                 },
-              },
-            });
+              });
 
-          await new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'list')
-            .then(() => { expect(true).toStrictEqual(true); });
+            await new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'browseEnabled', { userId: 1 })
+              .then(() => { expect(true).toStrictEqual(true); });
```
VincentMolinie, line 35: You are using `await` so get rid of the `then` it will be much easier to read

test/services/permissions.test.js
```diff
@@ -9,154 +9,164 @@ describe('services > permissions', () => {
...
+                      list: true,
+                    },
+                  },
+                },
+              });
 
-        await expect(new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'list'))
-          .rejects.toThrow("'list' access forbidden on Users");
+            await new PermissionsChecker('envSecret', 2).checkPermissions('Users', 'browseEnabled', { userId: 1 })
+              .then(() => { expect(true).toStrictEqual(true); });
```
VincentMolinie, line 131: Looking at your test you should use `.not.toThrow()` which is way more explicit than `expect(true).toStrictEqual(true)` 😅

test/services/permissions.test.js
```diff
@@ -404,257 +559,1036 @@ describe('services > permissions', () => {
...
+                },
+              });
+
+              const smartActionParameters = {
+                actionName: 'save',
+                userId: '1',
+              };
+
+              const result = await new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'actions', smartActionParameters);
+              expect(result).toBeUndefined();
```
VincentMolinie, line 1138: What is that test 😅 ? Are you testing that it does not throw ?

VincentMolinie (changes requested):
Nice that you implemented all those tests but it's way to hard to follow 😅 ... It would have been better to just implement Unit test so it's easier to read and understand ^^

src/services/permissions-checker.js
```diff
@@ -16,58 +16,144 @@ class PermissionsChecker {
 
   static expirationInSeconds = EXPIRATION_IN_SECONDS;
 
-  // This permissionsPerRendering is the cache, shared by all instances of PermissionsChecker.
-  static permissionsPerRendering = {};
+  static isrolesACLActivated;
```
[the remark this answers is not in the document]
adriguy: Good catch!! 🙏

src/services/permissions-checker.js
adriguy: Not sure why it will contain the old format 🤔 
But sure it would be better to user a new object here! 🙏

src/services/permissions-checker.js
adriguy: Yes sorry, hopefully you don't mind!

src/services/permissions-checker.js
adriguy: 😄  OMG that's true thanks!!

test/services/permissions.test.js
adriguy: This is old code but you're right

test/services/permissions.test.js
adriguy: Yes sorry I copied it from the existing tests... I guess it is just checking that the permission is allowed, so that it doesn't throw indeed!
I'll change them all.

src/services/permissions-checker.js
adriguy: OMG thanks!

src/services/permissions-cache.js
```diff
@@ -0,0 +1,168 @@
...
+    if (PermissionsCache.isRolesACLActivated) {
+      return collectionsPermissions ? collectionsPermissions.data : null;
+    }
+    return renderingPermissions ? renderingPermissions.data : null;
+  }
+
+  static getScopePermissions(renderingId, collectionName) {
+    const renderingPermissions = PermissionsCache._getRenderingPermissions(renderingId);
+    if (renderingPermissions && renderingPermissions.data
+      && renderingPermissions.data[collectionName]) {
```
VincentMolinie, line 45:
```suggestion
    if (renderingPermissions
      && renderingPermissions.data
      && renderingPermissions.data[collectionName]) {
```

src/services/permissions-cache.js
```diff
@@ -0,0 +1,168 @@
...
+      ? PermissionsCache.transformPermissionsFromOldToNewFormat(permissions)
+      : null;
+
+    PermissionsCache.permissions.renderings = {
+      ...PermissionsCache.permissions.renderings,
+      [renderingId]: {
+        data: newFormatPermissions,
+        lastRetrieve: moment(),
+      },
+    };
```
VincentMolinie, line 104: While doing a copy while you can just do 
```js
if (!PermissionsCache.permissions.renderings) {
  PermissionsCache.permissions.renderings = {}
}
PermissionsCache.permissions.renderings[renderingId] = { data: ... };
```

src/services/permissions-cache.js
```diff
@@ -0,0 +1,168 @@
...
+      data: permissions.collections,
+      lastRetrieve: moment(),
+    };
+    PermissionsCache.permissions.renderings = {
+      ...PermissionsCache.permissions.renderings,
+      [renderingId]: {
+        data: permissions.renderings ? permissions.renderings[renderingId] : null,
+        lastRetrieve: moment(),
+      },
+    };
```
VincentMolinie, line 118: Same Here

src/services/permissions-cache.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (!lastRetrieve) {
+      return true;
+    }
+
+    const elapsedSeconds = currentTime.diff(lastRetrieve, 'seconds');
+    return elapsedSeconds >= PermissionsCache.expirationInSeconds;
+  }
+
+  static async retrievePermissions(environmentSecret, renderingId) {
```
VincentMolinie, line 162: As the function to retrieve the permissions is here plus there is also all the function. to get the permissions I think this service should be called `PermissionsGetter`

src/services/permissions-checker.js
```diff
@@ -304,7 +158,7 @@ class PermissionsChecker {
   async checkPermissions(collectionName, permissionName, permissionInfos) {
     // TODO IN NEXT PR: Distinguish collectionsPermissionExpired and scopePermissionExpired for
     // rolesACL format and retrieve only scope when needed.
-    if (this._isPermissionExpired(permissionName)) {
+    if (PermissionsCache.isPermissionExpired(this.renderingId, permissionName)) {
```
VincentMolinie, line 161: Well the `isPermissionExpired` should actually be a private function. And here you should have only one function to get the permissions
```js
const permissions = await PermissionsGetter.getPermissionsForRendering(this.renderingId);
```

VincentMolinie: This get permissions should return **ALWAYS** the same format and normally you should have no need to call `PermissionsGetter` anywhere else in this file 😉

src/services/permissions-checker.js
```diff
@@ -241,7 +119,7 @@ class PermissionsChecker {
   }
 
   async _isAllowed(collectionName, permissionName, permissionInfos) {
-    const collectionsPermissions = PermissionsChecker.getCollectionsPermissions(this.renderingId);
+    const collectionsPermissions = PermissionsCache.getPermissions(this.renderingId);
```
VincentMolinie, line 122: Like I said the permissions should be retrieved only once

src/services/permissions-checker.js
```diff
@@ -259,7 +137,7 @@ class PermissionsChecker {
         return PermissionsChecker._isCollectionBrowseAllowed(
           collectionsPermissions[collectionName],
           permissionInfos,
-          this._getScopePermissions(collectionName),
+          PermissionsCache.getScopePermissions(this.renderingId, collectionName),
```
VincentMolinie, line 140: Same here

src/services/permissions-checker.js
```diff
@@ -269,32 +147,8 @@ class PermissionsChecker {
...
-      return true;
-    }
-
-    const elapsedSeconds = currentTime.diff(lastRetrieve, 'seconds');
-    return elapsedSeconds >= PermissionsChecker.expirationInSeconds;
-  }
-
   async _retrievePermissionsAndCheckAllowed(collectionName, permissionName, permissionInfos) {
-    await this._retrievePermissions();
+    await PermissionsCache.retrievePermissions(this.environmentSecret, this.renderingId);
```
VincentMolinie, line 151: Not the job of this service. The `PermissionsGetter` should decide by itself if it should retrieve the permissions or not

test/services/permissions.test.js
```diff
@@ -101,8 +90,8 @@ describe('services > permissions', () => {
                 },
               });
 
-            await new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'browseEnabled', { userId: 1 })
-              .then(() => { expect(true).toStrictEqual(true); });
+            expect(() => new PermissionsChecker('envSecret', 1).checkPermissions('Users', 'browseEnabled', { userId: 1 }))
+              .not.toThrow();
```
VincentMolinie, line 94: I'm pretty sure this is not the right syntax 🤔 
```suggestion
const permissionsChecker = new PermissionsChecker('envSecret', 1);
await expect(permissionsChecker.checkPermissions('Users', 'browseEnabled', { userId: 1 })).not.toThrow();
```

test/services/permissions.test.js
```diff
@@ -127,8 +116,8 @@ describe('services > permissions', () => {
                 },
               });
 
-            await new PermissionsChecker('envSecret', 2).checkPermissions('Users', 'browseEnabled', { userId: 1 })
-              .then(() => { expect(true).toStrictEqual(true); });
+            expect(() => new PermissionsChecker('envSecret', 2).checkPermissions('Users', 'browseEnabled', { userId: 1 }))
```
VincentMolinie, line 119: Same here and etc...

VincentMolinie (changes requested):

src/services/permissions-checker.js
adriguy: Well actually no... because the permissions retrieval will depend on whether the cached permission is allowed or not!
So we'll need one service that will interfere with the other.
I left the retrievePermissions call in the permissions checker for now, let's discuss about it when possible.

src/services/permissions-checker.js
adriguy: Ok for the same format of the `PermissionsGetter.getPermissions`.
However, `isPermissionExpired` should not be a private function to me (see my comment below).

src/services/permissions-getter.js
```diff
@@ -0,0 +1,193 @@
...
+
+  static getLastRetrieveTime(renderingId, permissionName) {
+    // In the case of rolesACL format and browseEnabled permission, the last retrieve to be taken
+    // into account is the one stored by rendering (because of the scope information).
+    if (PermissionsGetter.isRolesACLActivated && PermissionsGetter._getPermissionsInCollections() && permissionName !== 'browseEnabled') {
+      return PermissionsGetter._getPermissionsInCollections().lastRetrieve;
+    }
+    return PermissionsGetter._getPermissionsInRendering(renderingId)
+      ? PermissionsGetter._getPermissionsInRendering(renderingId).lastRetrieve
+      : null;
```
VincentMolinie, line 155: You have an issue ere, but not a big one. I'll talk about it in your next PR

VincentMolinie (changes requested):

src/services/permissions-getter.js
adriguy: This function will disappear in the next PR anyway 😉  But sure let's talk about it whenever you want!

adriguy: Oh I think I know what you mean. It concerns the "browseEnabled" case when you can have retrieve only on renderings?
If it is, don't worry I take care of this in the next PR 👍 
In this PR, for rolesACL, the lastRetrieve are the same in collectionPermissions and in renderingPermissions.

VincentMolinie (approved):
Good Job !

PR #574, opened 2020-12-02
fix(related-data): use same reference on record for dataValues and direct attributes

GuillaumeCisco:
## Pull Request checklist:

For not dealing with too much modification, I choose to directly overwrite the magic accessor on a field if one of its smart field has the same name.
As we are in a serialization process, this will not introduce side effects.
Not using dataValues (which in sequelize allow to isolate a context without magic accessors) will make it works with forest-mongoose too.

Others options are:
- use a dedicated smartValues context in record: https://github.com/ForestAdmin/forest-express/pull/443
- rework the code for injecting smart fields after json serialization


src/services/smart-fields-values-injector.js
```diff
@@ -68,9 +76,10 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
 
   this.perform = () =>
     P.each(schema.fields, (field) => {
-      if (!record[field.field]) {
+      if (record
+          && !Object.prototype.hasOwnProperty.call(record, field.field)) {
```
GuillaumeCisco, line 80: This is the real fix.
You can follow this [discussion](https://github.com/ForestAdmin/forest-express/pull/470/files#r466433245) for more details.

SteveBunlon: 👍

jeffladiray (commented):
Peer reviewed with Steve, but I think he will also add a few comments.

Manual tests on sequelize seems ok to me. Added a few comments/suggestion, but feature-wise it LGTM.

test/services/smart-fields-values-injector.test.js
```diff
@@ -1,53 +1,100 @@
+// eslint-disable-next-line max-classes-per-file
```
SteveBunlon, line 1: [OPINION] We can remove this line if we something else than classes. I think that using simple objects could be enough

test/services/smart-fields-values-injector.test.js
```diff
@@ -1,53 +1,100 @@
...
-      expect(record.user).not.toBeUndefined();
+      expect(addressRecord.smart_user).not.toBeUndefined();
+      expect(addressRecord.smart_user.smart).toStrictEqual({ foo: 'bar' });
+      expect(addressRecord.user.smart).toStrictEqual({ foo: 'bar' });
     });
-    it('should inject the Smart Field of the record referenced by the Smart Relationship', async () => {
-      expect.assertions(1);
+  });
+
+  describe('with a Smart Relationship that reference a collection having a Smart Field whose name is a magic accessor', () => {
```
SteveBunlon, line 84:
```suggestion
  describe('with a Smart Relationship that references a collection having a Smart Field whose name is a magic accessor', () => {
```

SteveBunlon (changes requested):
Looks good to me, I've tested it using mongoose (smart field + smart relationship). 

I've written some comments but most of them are [OPINION] so skippable.

src/services/smart-fields-values-injector.js
GuillaumeCisco: > I've tested it yet, but by reading this line I'm guessing that magic accessors are not considered as own properties. Correct ?

absolutely right sir!

src/services/smart-fields-values-injector.js
```diff
@@ -90,8 +99,13 @@ function SmartFieldsValuesInjector(record, modelName, fieldsPerModel, depth = 0)
               return null;
             }
 
-            if (!record[field.field][fieldAssociation.field]
-              && (fieldAssociation.get || fieldAssociation.value)) {
+            if (record
+                && record[field.field]
+                && !Object.prototype.hasOwnProperty.call(
```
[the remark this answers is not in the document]
GuillaumeCisco: interesting, will try :)

test/services/smart-fields-values-injector.test.js
GuillaumeCisco: I have to use class because I need to update its prototype.
Modifying the prototype of `Object` could lead to prototype pollution, that's why I'd like to keep it away :)

test/services/smart-fields-values-injector.test.js
```diff
@@ -1,53 +1,100 @@
+// eslint-disable-next-line max-classes-per-file
 const _ = require('lodash');
+
```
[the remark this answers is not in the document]
GuillaumeCisco: just a question, don't we separate import from third parties and import from our own code?

src/services/smart-fields-values-injector.js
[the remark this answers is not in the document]
GuillaumeCisco: Unfortunately, it will fail.
First arg need to be a no null `Object`

test/services/smart-fields-values-injector.test.js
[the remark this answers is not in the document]
SteveBunlon: We don't, this is not necessary then. Personally I like to compact imports and this is how it's done through the app

test/services/smart-fields-values-injector.test.js
SteveBunlon: Okay, nice and clear

SteveBunlon (approved):

jeffladiray (approved):
LGTM

PR #584, opened 2020-12-07
fix(smart-actions): reset value when not present in enums in hook response

rap2hpoutre:
## Pull Request checklist:

### DIFF
--- a/src/services/smart-action-hook.js
+++ b/src/services/smart-action-hook.js
@@ -43,8 +43,14 @@ class SmartActionHook {
       throw new Error('fields must be unchanged (no addition nor deletion allowed)');
     }
 
-    // Apply result on fields (transform the object back to an array), preserve order.
-    return fields.map((field) => result[field.field]);
+    return fields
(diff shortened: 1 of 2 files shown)


src/services/smart-action-hook.js
```diff
@@ -43,8 +43,14 @@ class SmartActionHook {
...
-    // Apply result on fields (transform the object back to an array), preserve order.
-    return fields.map((field) => result[field.field]);
+    return fields
+      // Apply result on fields (transform the object back to an array), preserve order.
+      .map((field) => result[field.field])
+      // Reset `value` when not present in `enums` (which means `enums` has changed).
+      .map((field) => {
+        if (!Array.isArray(field.enums) || field.enums.includes(field.value)) return field;
+        return { ...field, value: null };
+      });
```
GuillaumeCisco, line 53: We should avoid double `map`.
Can we do this in one loop?

rap2hpoutre: ok

GuillaumeCisco (approved):
Everything is alright to me 👍

PR #614, opened 2021-01-13
fix(forestadmin-schema): regenerate forestadmin schema only when files are valid

jeffladiray:
## Pull Request checklist:

CC complains about diff-coverage being not sufficient, but I don't think I'll be able to test the `map: xxx` code without reworking the `index.js` file. Using the real `require-all` module instead of a mock implies that I need to have a whole directory with files, which should contains syntax error in order to test the behavior I added.

If you find have any good idea on how I can test this function without loads of fixtures...

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -63,6 +63,7 @@ function getModels() {
 
 function requireAllModels(modelsDir) {
   if (modelsDir) {
+    let lastFileLoadedPath;
     try {
       const isJavascriptOrTypescriptFileName = (fileName) =>
         fileName.endsWith('.js') || (fileName.endsWith('.ts') && !fileName.endsWith('.d.ts'));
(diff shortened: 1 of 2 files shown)


SteveBunlon (approved):
🔥

SteveBunlon (dismissed):

src/index.js
```diff
@@ -75,10 +76,20 @@ function requireAllModels(modelsDir) {
         excludeDirs: /^__tests__$/,
         filter: (fileName) =>
           isJavascriptOrTypescriptFileName(fileName) && !isTestFileName(fileName),
+        map: (name, filePath) => {
+          lastFileLoadedPath = filePath;
+          return name;
+        },
```
jeffladiray, line 82: I can't find a way of testing this without adding loads of unrelated fixtures. Structure & behavior of `index.js` highly relies on `fs` & project structure, and injecting `forest-express-sequelize/mongoose` within `forest-express` does not seems to be a good idea.

`require-all` is also not used everywhere in our code (Like [here](https://github.com/ForestAdmin/forest-express/blob/42fffaf52541aaa5edc9df33fb0645ec389bedbd/src/integrations/index.js#L6) ), so a solution using `mock-fs` does not work without adding a lot of extra mock data.

Moreover, this function is not exposed, thus impossible to unit test properly.

Any idea ?

arnaud-moncel (approved):
LGTM i tested it manually !

SteveBunlon:
Works like charm 🧙

SteveBunlon (approved):

PR #615, opened 2021-01-15
ci: move from travis to github action

VincentMolinie:
## Pull Request checklist:


.github/workflows/build.yml
```diff
@@ -0,0 +1,84 @@
...
+        env:
+          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+        with:
+          coverageCommand: yarn test:coverage
+
+  deploy:
+    name: Release package
+    runs-on: ubuntu-latest
+    needs: [test]
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/beta')
```
arnaudbesnier, line 60: Will there be 2 releases without the "[skip ci]" condition?

.github/workflows/build.yml
```diff
@@ -0,0 +1,84 @@
+name: Build, Test and Deploy
+
+on:
+  push:
+    branches:
+      - master
+      - beta
+  pull_request:
+
+
```
arnaudbesnier, line 10: [Style] Can we get rid of this double newline?

.github/workflows/build.yml
VincentMolinie: Nop because test is executed only if there is no `[skip ci]` so as deploy `needs` test it won't start.

VincentMolinie: I understand why you said that the if on test is actually not necessary it's just in case one day someone as the idea to launch them in parallel but I can remove it if you want

arnaudbesnier (approved):

PR #622, opened 2021-01-25
feat(authentication): return a specific error during authorization if received an error about 2FA

ghusse:
This modification is needed to identify why the authorization failed, and display the correct error message in the frontend.

In this case, instead of a 500 with an error message, we get a 403 and the error's name correctly encoded in the response, just as the server sends errors.

Linked to CU-ccukn4

## Pull Request checklist:

### DIFF
--- a/src/services/authorization-finder.js
+++ b/src/services/authorization-finder.js
@@ -30,9 +30,8 @@ class AuthorizationFinder {
     switch (error.message) {
       case this.errorMessages.SERVER_TRANSACTION.SECRET_AND_RENDERINGID_INCONSISTENT:
         return new InconsistentSecretAndRenderingError();
-      case this.errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND: {
+      case this.errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND:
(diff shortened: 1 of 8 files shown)


src/services/authorization-finder.js
```diff
@@ -22,20 +26,28 @@ class AuthorizationFinder {
...
-        break;
-      case this.errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND:
-        errorMessageToForward = 'Cannot retrieve the project you\'re trying to unlock. '
-            + 'Please check that you\'re using the right environment secret regarding your project and environment.';
-        break;
-      default: errorMessageToForward = undefined;
+        return new InconsistentSecretAndRenderingError();
+      case this.errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND: {
+        return new SecretNotFoundError();
+      }
```
DrRaider, line 35:
```suggestion
      case this.errorMessages.SERVER_TRANSACTION.SECRET_NOT_FOUND:
        return new SecretNotFoundError();
```

DrRaider (changes requested):

DrRaider (approved):

PR #688, opened 2021-04-15
build: add source-maps to debug build

romain-gilliotte:
This PR makes the dev experience it a bit easier

- Make "watch" builds faster by dropping the `onchange` devDependency and watching the src folder with babel instead. This allows recompiling only files which changes, instead of recompiling the whole project at each change which takes seconds at each change.
- Enable source-maps, which allow using vscode debugger & breakpoints
- Exclude the dist folder from eslint configuration, to prevent eslint from flooding vscode "Problems" tab.


package.json
```diff
@@ -102,4 +101,4 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
   }
-}
+}
```
arnaudbesnier, line 104: I guess your editor should automate the newline at the end of files to be consistent with our codebase.

package.json
```diff
@@ -88,8 +87,8 @@
     "supertest": "4.0.2"
   },
   "scripts": {
-    "build": "./node_modules/.bin/babel src -d dist && echo '\n\\033[0;34m[+] \\033[0;32mBuild done\\033[0m'",
-    "build:watch": "onchange 'src/**/*.js' -i -- yarn build",
+    "build": "babel src -d dist",
+    "build:watch": "babel --source-maps inline --watch --out-dir dist src",
```
arnaudbesnier, line 91: Looks nice ✨

package.json
```diff
@@ -88,8 +87,8 @@
     "supertest": "4.0.2"
   },
   "scripts": {
-    "build": "./node_modules/.bin/babel src -d dist && echo '\n\\033[0;34m[+] \\033[0;32mBuild done\\033[0m'",
-    "build:watch": "onchange 'src/**/*.js' -i -- yarn build",
+    "build": "babel src -d dist",
```
arnaudbesnier, line 90: I have 2 questions:
- Why do you want to get rid of the `echo`?
- I have never dug into this topic. I don't know why some projects use relative path to use dependencies and why some don't. I actually think it is inconsistent in our own codebase and I know some developers favored the relative path `./node_modules/...` maybe to prevent dependencies conflicts ?

package.json
romain-gilliotte: resolved

package.json
romain-gilliotte: About the echo, because babel already outputs a message to tell the build is done, so two of them are displayed
```console
[+] Build done
✨  Done in 3.69s.
```

About removing `node_modules/.bin`, npm/yarn is supposed to put that folder in the PATH anyways, so I don't think it matters.

I had no reason for removing it, I just wanted both build commands to be consistent (which I did a very bad job at: eslint still has the prefix, which jest does not).

romain-gilliotte: Also the echo command prevents the users from adding parameters to babel, which is convenient.

For instance: if you want coloring on the output of babel
```
yarn build --highlightCode
```

larcin (approved):

PR #695, opened 2021-04-26
feat(scopes): add scopes retrieval service

larcin:
Usage: 

```
import { scopeManager } from 'forest-express';

// With user being the one from the request and `environments` being the collectionName
const scopes = await scopeManager.getScopeForUser(user, 'environments');
```
Format (only collections having scopes are retrieved):
```
{
  aggregator: 'and',
  conditions: [
    {
      field: 'type',
      operator: 'equal',
      value: 'production',
    },
  ],
}
```

## Pull Request checklist:


src/services/exposed/scope-service.js
```diff
@@ -0,0 +1,13 @@
...
+
+class ScopeService {
+  static async getScopes(renderingId) {
+    const { configStore, forestServerRequester } = context.inject();
+    const { envSecret } = configStore.lianaOptions;
+    const queryParams = { renderingId };
+
+    return forestServerRequester.perform('/liana/scopes', envSecret, queryParams);
+  }
+}
```
romain-gilliotte, line 11: In order to complete the feature, this service should be able to provide a filter which has the same format as the filters which are used for segments, charts, and so on.

This service should perform the substitution of $currentUser.* entries

```
// this._model == reviews sequelize model
// user == user from request.user (it contains the renderingId)
await getScope(this._model, user) == {
   "aggregator":"and",
   "conditions":[
       {"field":"book:id","operator":"equal","value":2},
       {"field":"id","operator":"equal","value":3}
    ]
}
```

romain-gilliotte (changes requested):
See comments in files.

+ tests are missing :)

src/services/exposed/scope-service.js
```diff
@@ -0,0 +1,18 @@
+const context = require('../../context');
+
+class ScopeService {
+  static async getScopes(renderingId) {
+    const { configStore, forestServerRequester } = context.inject();
+    const { envSecret } = configStore.lianaOptions;
```
romain-gilliotte, line 6: checkDependency?

src/services/exposed/scope-service.js
```diff
@@ -0,0 +1,18 @@
+const context = require('../../context');
```
romain-gilliotte, line 1: import from?

src/services/exposed/scope-service.js
larcin: FES and FE have no `checkDependencies` implemented for now. It could be great to add it but a bit irrelevant through that PR

romain-gilliotte (approved):

src/services/scope-manager.js
```diff
@@ -0,0 +1,76 @@
...
+
+  async _fetchScopes(renderingId) {
+    const { envSecret } = this.configStore.lianaOptions;
+    const queryParams = { renderingId };
+
+    return this.forestServerRequester.perform('/liana/scopes', envSecret, queryParams);
+  }
+
+  async _refreshScopesCache(renderingId) {
+    const scopes = await this._fetchScopes(renderingId);
```
romain-gilliotte, line 60: there is a race condition here.

if the cache is outdated, and the method is called many times in a short burst, the query will be triggered many times.
You should update the time in the cache before, then make the query

src/services/scope-manager.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  static _formatDynamicValues(userId, collectionScope) {
+    if (!collectionScope?.scope?.filter) return null;
+
+    collectionScope.scope.filter.conditions.forEach((condition) => {
+      if (condition.value
+        && `${condition.value}`.startsWith('$')
+        && collectionScope.scope.dynamicScopesValues.users[userId]) {
+        condition.value = collectionScope
```
romain-gilliotte, line 30: Are you modifying the entry from the cache without cloning it?
That won't work if multiple users call the service

romain-gilliotte (changes requested):

src/services/scope-manager.js
```diff
@@ -0,0 +1,93 @@
...
+          .scope
+          .dynamicScopesValues
+          .users[userId][condition.value];
+      }
+    });
+
+    return collectionScopeCopy.scope.filter;
```
romain-gilliotte, line 41: You are cloning the big JSON, which contains data about all users just to perform replacements in the filter.
It works, but may be slow.

Did you know about `cloneDeepWith` in lodash?
You can perform substitution without knowing details about the internals of the object.

```suggestion
  return _.cloneDeepWith(collectionScope?.scope?.filter, item => (
    typeof item === 'string' && item.startsWith('$currentUser')
      ? collectionScope.scope.dynamicScopesValues.users[userId][item]
      : undefined
  ));
```

romain-gilliotte (changes requested):

romain-gilliotte (approved):

PR #700, opened 2021-04-29
feat(scopes)!: stop checking that scopes are properly provided as filters

romain-gilliotte:
Breaking Change !!


src/services/permissions-checker.js
```diff
@@ -43,105 +40,20 @@ class PermissionsChecker {
...
-    if (!scope) return true;
-
-    try {
-      return this._isScopeValid(permissionInfos, scope);
-    } catch (error) {
-      this.logger.error(error);
-      return false;
-    }
+  // eslint-disable-next-line class-methods-use-this
+  async _isCollectionBrowseAllowed(collectionPermissions, permissionInfos) {
```
larcin, line 44: [MISC] Can't you make it a static method and call it as the others: `PermissionsChecker._isCollectionBrowseAllowed`?

(that would dismiss the eslint rule removal)

larcin (changes requested):

larcin (approved):

PR #711, opened 2021-05-12
feat: ensure smart action target recordIds are in scopes

larcin:
## Pull Request checklist:


src/middlewares/permissions.js
```diff
@@ -86,6 +88,50 @@ class PermissionMiddlewareCreator {
...
+      const { idField } = Schemas.schemas[modelName];
+
+      // TODO: scope smartAction calls properly on table with composite primary keys
+      if (idField === 'forestCompositePrimary') return next();
+
+      const tragetRecordIds = attributes.ids;
+      const checkIdsFilter = JSON.stringify({
+        field: idField,
+        operator: 'in',
+        value: tragetRecordIds,
```
romain-gilliotte, line 118: traget -> target

src/middlewares/permissions.js
```diff
@@ -86,6 +88,50 @@ class PermissionMiddlewareCreator {
...
+      const attributes = PermissionMiddlewareCreator._getRequestAttributes(request);
+
+      // if performing a `selectAll` let the `getIdsFromRequest` handle the scopes
+      if (attributes.allRecords) return next();
+
+      const modelName = this.configStore.Implementation.getModelName(model);
+      const { idField } = Schemas.schemas[modelName];
+
+      // TODO: scope smartAction calls properly on table with composite primary keys
+      if (idField === 'forestCompositePrimary') return next();
```
romain-gilliotte, line 112: Maybe we should debate with Arnaud if it's OK to allow this.
The use case is B2B2B, security matters much more than with all previous forest admin features.

If two customers which don't know each other are using forest admin, we really don't want information to leak between them. 

How about:
```
return response.status(400).send({ error: 'Smart Action are not supported on models with composite primary keys' });
```

src/middlewares/permissions.js
```diff
@@ -15,10 +17,10 @@ class PermissionMiddlewareCreator {
   }
 
   _getSmartActionInfoFromRequest(request) {
-    const smartActionEndpoint = request.originalUrl;
+    const smartActionEndpoint = `${request.baseUrl}${request.path}`;
     const smartActionHTTPMethod = request.method;
     const smartAction = Schemas.schemas[this.collectionName].actions.find((action) => {
-      const endpoint = action.endpoint || `/forest/actions/${parameterize(action.name)}`;
+      const endpoint = action.endpoint || `/actions/${parameterize(action.name)}`;
```
romain-gilliotte, line 23: Why remove the `/forest/` now that we're using `${request.baseUrl}${request.path}`?

src/middlewares/permissions.js
romain-gilliotte: Just to be clear, the code is OK.
It's just that I think that it is better if arnaud knows about the security issue before we merge this.

This way, he won't be surprised if the next security audit finds it

src/middlewares/permissions.js
```diff
@@ -86,6 +88,50 @@ class PermissionMiddlewareCreator {
...
+      const tragetRecordIds = attributes.ids;
+      const checkIdsFilter = JSON.stringify({
+        field: idField,
+        operator: 'in',
+        value: tragetRecordIds,
+      });
+
+      // count records matching the provided filters (with scopes applied by the RecordCounter)
+      const count = await new RecordsCounter(
+        model, request.user, { filters: checkIdsFilter, timezone: 'Europe/Paris' },
```
romain-gilliotte, line 123: This should be the timezone from the request.
Otherwise the frontend will break if using timezone on scopes (if you are in new york, record that are in-scope on the NY timezone might not be in the Paris one)

romain-gilliotte (changes requested):
I added some comments in the code

Note: I did not test functionality yet

src/middlewares/permissions.js
larcin: The idea here was to have a first version working for non-compositeKeys collection (as we're targeting a mongoose project first) and later on the project add the support for that

src/middlewares/permissions.js
larcin: True, I wanted to edit it later on but forgot to, I'll had it right away!

src/middlewares/permissions.js
larcin: I don't recall why I got rid of it, I'll add it back

src/middlewares/permissions.js
```diff
@@ -86,6 +88,50 @@ class PermissionMiddlewareCreator {
...
+  static _getRequestAttributes(request) {
+    const hasBodyAttributes = request.body && request.body.data && request.body.data.attributes;
+    return hasBodyAttributes
+      && new QueryDeserializer(request.body.data.attributes).perform();
+  }
+
+  // generate a middleware that will check that ids provided by the request exist
+  // whithin the registered scope
+  _ensureRecordIdsInScope(model) {
+    if (!model) throw new Error('missing model');
```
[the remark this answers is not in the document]
larcin: This will be used by both `forest-expresss-sequelize` and `forest-expresss-mongoose`, thus what about `missing model in the smartAction middleware definition`, a quick check on either the documentation or the support should provide more details?

romain-gilliotte (approved):
👍

PR #717, opened 2021-05-19
feat(smart-action): user can add dynamically field with hook on smart action form

arnaud-moncel:
BREAKING CHANGE: change hook is no longer choosen by the field name, field need to have hook defined inside it definition by adding a props hook

## Pull Request checklist:


src/index.js
```diff
@@ -114,12 +115,15 @@ function generateAndSendSchema(envSecret) {
...
       collection.actions.forEach((action) => {
         if (!action.name) {
           logger.warn(`An unnamed Smart Action of collection "${collection.name}" has been ignored.`);
-        } else if (isFieldsInvalid(action)) {
-          logger.error(`Cannot find the fields you defined for the Smart action "${action.name}" of your "${collection.name}" collection. The fields option must be an array.`);
+        } else {
+          try {
+            smartActionFieldValidator.validateSmartActionFields(action, collection.name);
+          } catch (error) {
+            logger.error(error.message);
```
jeffladiray, line 125: I've distinguish on the rails side warn (For invalid property that should still work) and error for things that might be broken. Maybe we should do the same here ?

src/services/smart-action-field-validator.js
```diff
@@ -32,27 +32,47 @@ class SmartActionFieldValidator {
       type,
     } = field;
 
-    if (!fieldName) throw new Error('field attribute must be defined.');
+    if (!fieldName) throw new Error(`field attribute inside fileds array on the smart action "${actionName}" must be defined.`);
```
jeffladiray, line 35:
```suggestion
    if (!fieldName) throw new Error(`field attribute inside fields array on the smart action "${actionName}" must be defined.`);
```

src/services/smart-action-field-validator.js
```diff
@@ -32,27 +32,47 @@ class SmartActionFieldValidator {
...
 
-    if (typeof fieldName !== 'string') throw new Error('field attribute must be a string.');
+    if (typeof fieldName !== 'string') throw new Error(`field attribute inside fileds array on the smart action "${actionName}" must be a string.`);
```
jeffladiray, line 37:
```suggestion
    if (typeof fieldName !== 'string') throw new Error(`field attribute inside fields array on the smart action "${actionName}" must be a string.`);
```

test/services/smart-action-field-validator.test.js
```diff
@@ -50,73 +54,171 @@ describe('services > smart-action-field-validator', () => {
...
+      const actionName = 'actionTest';
+      const hooks = {
+        onChange: () => {},
+      };
+
+      expect(() => smartActionFieldValidator.validateFieldChangeHook(field, actionName, hooks))
+        .not.toThrow();
+    });
+
+    it('should throw if field have not defined hook', () => {
```
jeffladiray, line 160:
```suggestion
    it('should throw if field have undefined hook', () => {
```

jeffladiray, line 160: [Opinion] Labels for these tests cases aren't obvious to me

`should not throw if field does not have defined hook` => `when the field does not use the change hook feature`.`it should do nothing`
`should not throw if field have defined hook` => `when the field uses the hook feature`.`when the hook is correctly defined`.`it should do nothing`
`should throw if field have not defined hook` => `when the field uses the hook feature`.`when the field hook does not exist`.`it should throw an error` 

Would be clearer to me

test/services/smart-action-field-validator.test.js
```diff
@@ -50,73 +54,171 @@ describe('services > smart-action-field-validator', () => {
...
+      const actionName = 'actionTest';
+      const hooks = {};
+
+      expect(() => smartActionFieldValidator.validateFieldChangeHook(field, actionName, hooks))
+        .toThrow(`The hook "${field.hook}" of "${field.field}" field on the smart action "${actionName}" is not defined.`);
+    });
+  });
+
+  describe('validateSmartActionFields', () => {
+    it('should do nothing if action does not have field', () => {
```
jeffladiray, line 173: Same here

`when the action does not have any fields`.`it should do nothing`
`when the action uses incorrect parameters`.`it should throw an error for fields not being an array` 
...

test/services/smart-action-hook.test.js
```diff
@@ -73,21 +62,164 @@ describe('services > smart-action-hook', () => {
...
+            1,
+            {
+              fields,
+              record: {},
+              changedField: null,
+            },
+          );
+        });
+
+        it('should assign null value if no one is provided', async () => {
```
jeffladiray, line 101:
```suggestion
        it('should assign null value if none is provided', async () => {
```

jeffladiray: All these are tests cases for incorrect usage, maybe a big `describe` would ease the readability

test/services/smart-action-hook.test.js
```diff
@@ -73,21 +62,164 @@ describe('services > smart-action-hook', () => {
...
+          const fields = [field, addedField];
+
+          const hook = jest.fn(() => ([field, addedField]));
+
+          await expect(smartActionHook.getResponse(action, hook, fields, {}))
+            .rejects.toThrow(`The hook "${addedField.hook}" of "${addedField.field}" field on the smart action "${action.name}" is not defined.`);
+        });
+      });
+
+      it('should have one less filed', async () => {
```
jeffladiray, line 184:
```suggestion
      it('should have one less field', async () => {
```
+ I think you are missing a describe here 🤔

jeffladiray (changes requested):
LGTM, apart from a few typos and tests structure.

We're almost there 🐨 🥇

src/index.js
arnaud-moncel: 🤷

test/services/smart-action-field-validator.test.js
arnaud-moncel: 😍

test/services/smart-action-field-validator.test.js
arnaud-moncel: 😍

src/index.js
arnaud-moncel: If we want to do that i will do the work inside an another PR

jeffladiray (dismissed):

jeffladiray (approved):

PR #719, opened 2021-05-24
fix(schema): remove engine version to prevent engine conflicts in DWO

olesyakorovina:
## Pull Request checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ dist
 
 # Editor
 .vscode
-.idea
 
 # Tests
 .forestadmin-schema.json
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ dist
 
 # Editor
 .vscode
+.idea
 
 # Tests
 .forestadmin-schema.json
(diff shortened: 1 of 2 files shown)


SteveBunlon (approved):
Looks good to me, needs to be double checked with @arnaudbesnier

src/index.js
```diff
@@ -136,7 +136,6 @@ function generateAndSendSchema(envSecret) {
       liana: configStore.Implementation.getLianaName(),
       liana_version: configStore.Implementation.getLianaVersion(),
       engine: 'nodejs',
-      engine_version: process.versions && process.versions.node,
```
arnaudbesnier, line 139: This change looks a bit naive to me and won't solve the problem on the long term.
What if the `orm_version`, `database_type`, ... is different from one environment to another?
Those scenarios are as possible as a different engine version.

Partially removing the meta data won't solve the issue and will make Forest lose data.
If the issue you try to solve is real, you'll have to make a choice between `meta` section removal in the schema or `meta` isolation for `apimapHash` computation.

.gitignore
```diff
@@ -31,6 +31,7 @@ dist
 
 # Editor
 .vscode
+.idea
```
arnaudbesnier, line 34: Not related to the change description.

PR #723, opened 2021-05-25
feat: extend scope manager to allow patching pre-existing filters

romain-gilliotte:
## Pull Request checklist:

### DIFF
--- a/src/services/scope-manager.js
+++ b/src/services/scope-manager.js
@@ -15,6 +15,20 @@ class ScopeManager {
     this.scopesCache = {};
   }
 
+  async appendScopeForUser(previousFilter, user, collectionName) {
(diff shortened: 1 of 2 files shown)


src/services/scope-manager.js
```diff
@@ -15,14 +15,28 @@ class ScopeManager {
     this.scopesCache = {};
   }
 
+  async appendScopeForUser(previousFilter, user, collectionName) {
+    const newFilter = await this.getScopeForUser(user, collectionName, true);
+    const filters = [previousFilter, newFilter].filter(Boolean);
```
larcin, line 20: [Naming] can't we find better names for `previous` and `new` filters? something like `existingFilter` and `scopeFilter` maybe?

PR #731, opened 2021-06-03
fix: correctly get the IP address from the request headers

olesyakorovina:
## Pull Request checklist:


SteveBunlon (approved):
Side note: now this code will be placed in `forest-express`, we can use it directly in forestadmin-server as it's actually a duplicates

PR #742, opened 2021-06-14
fix(intercom): fix date conversion from unix timestamp to js date of intercom attributes

### LINKED ISSUE #640: Intercom: dates are all in 1970
## Expected behavior

Dates shown in "Intercom attributes" are expected to be correct.

## Actual behavior

Dates look exactly like in your screenshot at https://docs.forestadmin.com/documentation/reference-guide/integrations#intercom - they're all in 1970.

Dates returned by the Intercom integration are returned as a Unix time, e.g.`1613415151`.
The Date constructor in JS is expecting a number of milliseconds since Unix epoch, e.g. `1613415151000`
```
> new Date(1613415151)
1970-01-19T16:10:15.151Z
```

## Context

* intercom-client: 2.11.0, used with Intercom API 2.0
* forest-express: 7.9.6
* express: 4.17.1

jeffladiray:
Close #640

## Pull Request checklist:

### DIFF
--- a/src/integrations/intercom/serializers/intercom-attributes.js
+++ b/src/integrations/intercom/serializers/intercom-attributes.js
@@ -3,6 +3,17 @@ const JSONAPISerializer = require('jsonapi-serializer').Serializer;
 function serializeIntercomAttributes(attributes, collectionName, meta) {
   const type = `${collectionName}_intercom_attributes`;
 
+  const unixTimestampToDateOrNull = (unixTimestamp) =>
+    unixTimestamp && new Date(unixTimestamp * 1000);
+
+  // Attributes keys ending with `_at` are unix timestamp
+  // thus they need to be converted to js Date
+  Object.entries(attributes).forEach(([attributeKey, attributeValue]) => {
+    if (attributeKey.endsWith('_at')) {
+      attributes[attributeKey] = unixTimestampToDateOrNull(attributeValue);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


src/integrations/intercom/serializers/intercom-attributes.js
```diff
@@ -3,6 +3,17 @@ const JSONAPISerializer = require('jsonapi-serializer').Serializer;
...
+    unixTimestamp && new Date(unixTimestamp * 1000);
+
+  // Attributes keys ending with `_at` are unix timestamp
+  // thus they need to be converted to js Date
+  Object.entries(attributes).forEach(([attributeKey, attributeValue]) => {
+    if (attributeKey.endsWith('_at')) {
+      attributes[attributeKey] = unixTimestampToDateOrNull(attributeValue);
+    }
+  });
+
```
GuillaumeCisco, line 16: No need to use `new Date(...)`, as only the timestamp can do the job.
Works well this way :)

GuillaumeCisco (approved):
Tested! 🏁 
Works great! 🎉

src/integrations/intercom/serializers/intercom-attributes.js
jeffladiray: I agree, but looking at other dates sent by the agents, it seems like they are sent as `new Date()` instead of timestamp, so I'll keep it this way for consistency

PR #753, opened 2021-06-23
feat: smart action hooks now have access to the http request

arnaud-moncel:
BREAKING CHANGE: record is no longer send to the hook midleware
values option on smart action is no longer supported

## Pull Request checklist:


jeffladiray (commented):
Shouldn't the commit message be a `feat`?

Hooks will now receive way more data than they did before this contribution, so maybe
`feat: smart action hooks now have access to the http request` would be a bit more explicit.

Otherwise, looks good works fine.

jeffladiray (approved):

PR #756, opened 2021-06-29
 fix: handle composite primary key case when checking permissions for smart actions

romain-gilliotte:
## Pull Request checklist:

### DIFF
--- a/src/middlewares/permissions.js
+++ b/src/middlewares/permissions.js
@@ -106,36 +106,41 @@ class PermissionMiddlewareCreator {
   // whithin the registered scope
   _ensureRecordIdsInScope() {
     return async (request, response, next) => {
-      const model = this.modelsManager.getModelByName(this.collectionName);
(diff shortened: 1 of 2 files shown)


src/middlewares/permissions.js
```diff
@@ -106,36 +106,41 @@ class PermissionMiddlewareCreator {
...
+        if (attributes.allRecords) {
+          return next();
+        }
+
+        // Otherwise, check that all records are within scope.
+        const { primaryKeys } = Schemas.schemas[this.collectionName];
+        const filters = JSON.stringify(primaryKeys.length === 1
+          ? { field: primaryKeys[0], operator: 'in', value: attributes.ids }
+          : {
+            aggregator: 'or',
```
romain-gilliotte, line 121: I don't understand why those lines show as untested in the code coverage
Do you get it?

larcin: I don't, my guess would be thatCC does not compute inline well ternaries like this one 🤷

larcin (approved):

PR #773, opened 2021-07-22
fix(actions): provide user and timezone to remove many action

romain-gilliotte:
user and timezone was not passed to remove many => it crashed
This is either
- a regression from scopes implementation
- failed merge?


adriguy (dismissed):
👌

adriguy (approved):

PR #793, opened 2021-08-30
chore: add a deprecation warning on the httpMethod property of smart actions

jeffladiray:
### General

### Security

- [x] Consider the security impact of the changes made


src/index.js
```diff
@@ -336,6 +336,12 @@ exports.init = async (Implementation) => {
   }
 };
 
+function getSmartActionsUsingHTTPMethod(actions) {
+  return actions.filter(
+    (action) => Object.hasOwnProperty.call(action, 'httpMethod'),
```
slimee, line 341: [Just opinion]
to do this I use simpler `action => action.httpMethod` since I'm sure httpMethod is not defined higher in prototype chain (maybe I'm wrong?)

slimee (dismissed):
🚀 
It works and it do the job.
Developpers will now have error/warning log message when their server starts.

src/index.js
jeffladiray: I'm not sure this can happen, but with my syntax, the log will also be displayed with `httpMethod: null` or `httpMethod: undefined`, which would not be the case with your syntax

slimee: ah! nice catch

slimee (approved):
🚀

PR #795, opened 2021-09-07
fix(smart field): fix filter on smart references

VincentMolinie:
### General

### Security


romain-gilliotte (approved):
- It works when used with the branch fix/smart-reference on sequelize ✅
- When I tried the smart reference case, I cannot inject includes
  - I made a request to fetch all ids of the collection I'm filtering that matched the condition for the "smart belongsto reference"
  - Then created a sequelize condition from this

Also!
I DID NOT TEST THAT THIS DOES NOT BREAKS THE MONGOOSE DRIVER
(as you are changing code in functions which are called by the "database drivers")

Are you releasing the mongoose patch soon? Otherwise, you'll need either to check or freeze the project to avoid any release that would bump the forest-express version in forest-express-mongoose

PR #879, opened 2022-04-15
fix(vulnerabilites): remove async dependency vulnerability

DrRaider:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)

### Security

You can find all the vulnerabilities [here](https://forest.slite.com/app/docs/wcubOYnJ6UBiNV)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "cors": "2.8.5",
     "csv-stringify": "1.0.4",
     "express": "4.17.1",
-    "express-jwt": "6.0.0",
+    "express-jwt": "https://github.com/auth0/express-jwt.git#c7881ad378063236d85b1e1b0f4a252b63b8e75b",
     "forest-ip-utils": "1.0.1",
     "http-errors": "1.7.3",
     "inflected": "2.0.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "superagent": "5.2.2",
     "uuid": "3.3.3",
     "verror": "1.10.0",
-    "winston": "3.2.1"
```
arnaudbesnier, line 57: you don’t need to update winston, only reinstall it to have its latest dependencies.

DrRaider: Looking at it's [changelog](https://github.com/winstonjs/winston/blob/master/CHANGELOG.md), I thought it would be a good idea to update it too. f you prefer I can rollback this change.

arnaudbesnier: separation of concerns, if you don’t need to upgrade it for the vulnerability purposes you should not do it.
You don’t want to add workload on top of this work (you never know, if it will work perfectly or not)

slimee (approved):
🚀 good job!
I was able to onboard a project with this branch.

PR #885, opened 2022-04-29
fix: log an error when smart action enums field contain null or undefined option

arnaud-moncel:
### General

- [X] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [X] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [X] Consider the security impact of the changes made

### DIFF
--- a/src/services/smart-action-field-validator.js
+++ b/src/services/smart-action-field-validator.js
@@ -38,7 +38,10 @@ class SmartActionFieldValidator {
 
     if (description && typeof description !== 'string') throw new Error(`description of "${fieldName}" on the smart action "${actionName}" must be a string.`);
 
-    if (enums && !Array.isArray(enums)) throw new Error(`enums of "${fieldName}" on the smart action "${actionName}" must be an array.`);
+    if (enums) {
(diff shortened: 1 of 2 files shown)


slimee (approved):
🚀

src/services/smart-action-field-validator.js
```diff
@@ -38,7 +38,10 @@ class SmartActionFieldValidator {
 
     if (description && typeof description !== 'string') throw new Error(`description of "${fieldName}" on the smart action "${actionName}" must be a string.`);
 
-    if (enums && !Array.isArray(enums)) throw new Error(`enums of "${fieldName}" on the smart action "${actionName}" must be an array.`);
+    if (enums) {
+      if (!Array.isArray(enums)) throw new Error(`enums of "${fieldName}" on the smart action "${actionName}" must be an array.`);
+      if (enums.some((option) => [null, undefined].includes(option))) throw new Error(`Invalid null or undefined option inside "${fieldName}" on the smart action "${actionName}".`);
```
romain-gilliotte, line 43: Would it work to go further and check that `typeof option !== 'string'` to prevent new bugs (I don't know if it is allowed to have number enums in forest, but I don't think so)

PR #951, opened 2022-10-18
refactor: support custom errors with data

Thenkei:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)

### Security

### DIFF
--- a/src/services/exposed/error-handler.js
+++ b/src/services/exposed/error-handler.js
@@ -13,16 +13,20 @@ function errorHandler({ logger } = {}) {
    */
   return function handleError(error, request, response, next) {
     if (error) {
+      const responseError = {
+        status: error.status || 500,
+        detail: error.message,
+        name: error.constructor.name,
+        ...(error.data ? { data: error.data } : {}),
+      };
+
       // NOTICE: Send the first error if any.
       if (error.errors && error.errors[0]) {
         if (error.errors[0].message) {
-          error.message = error.errors[0].message;
+          responseError.message = error.errors[0].message;
         }
         if (error.errors[0].name) {
-          error.name = error.errors[0].name;
-        }
-        if (error.errors[0].data) {
-          error.data = error.errors[0].data;
+          responseError.name = error.errors[0].name;
         }
       }
 
@@ -32,12 +36,7 @@ function errorHandler({ logger } = {}) {
       }
 
       response.status(error.status || 500).send({
-        errors: [{
-          status: error.status || 500,
-          detail: error.message,
-          name: error.name,
-          ...(error.data ? { data: error.data } : {}),
-        }],
+        errors: [responseError],
       });
     } else {
       next();
--- a/src/services/exposed/error-handler.js
+++ b/src/services/exposed/error-handler.js
@@ -14,12 +14,16 @@ function errorHandler({ logger } = {}) {
   return function handleError(error, request, response, next) {
     if (error) {
       // NOTICE: Send the first error if any.
-      if (error.errors && error.errors[0] && error.errors[0].message) {
-        error.message = error.errors[0].message;
-      }
-
-      if (error.errors && error.errors[0] && error.errors[0].name) {
-        error.name = error.errors[0].name;
+      if (error.errors && error.errors[0]) {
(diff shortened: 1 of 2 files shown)


src/services/exposed/error-handler.js
```diff
@@ -14,12 +14,16 @@ function errorHandler({ logger } = {}) {
   return function handleError(error, request, response, next) {
     if (error) {
       // NOTICE: Send the first error if any.
-      if (error.errors && error.errors[0] && error.errors[0].message) {
-        error.message = error.errors[0].message;
-      }
-
-      if (error.errors && error.errors[0] && error.errors[0].name) {
-        error.name = error.errors[0].name;
+      if (error.errors && error.errors[0]) {
```
Thenkei, line 17: I'm not sure to understand this use case on the agent side. Is it required due to `VError` usages?

src/services/exposed/error-handler.js
```diff
@@ -14,12 +14,16 @@ function errorHandler({ logger } = {}) {
...
+        if (error.errors[0].message) {
+          error.message = error.errors[0].message;
+        }
+        if (error.errors[0].name) {
+          error.name = error.errors[0].name;
+        }
+        if (error.errors[0].data) {
+          error.data = error.errors[0].data;
+        }
```
Thenkei, line 26: I this case this should be removed.

src/services/exposed/error-handler.js
```diff
@@ -32,6 +36,7 @@ function errorHandler({ logger } = {}) {
           status: error.status || 500,
           detail: error.message,
           name: error.name,
```
ghusse, line 38: Here you don't follow the same principle than on the `agent-nodejs` by using the constructor's name?

Thenkei: Well, I hadn't looked at all errors. It seems that all used `this.name = this.constructor.name;` syntax in constructor. So let's change it.

test/services/exposed/error-handler.test.js
```diff
@@ -1,5 +1,17 @@
 const sinon = require('sinon');
 const errorHandler = require('../../../src/services/exposed/error-handler');
+const ForbiddenError = require('../../../src/utils/errors/forbidden-error').default;
+
+class FakePayloadError extends ForbiddenError {
+  constructor(...args) {
+    super(...args);
+
+    this.name = this.constructor.name;
```
ghusse, line 9: Do you need this in this tests? I think it's using the constructor's name with your code.

test/services/exposed/error-handler.test.js
```diff
@@ -108,8 +120,37 @@ describe('services › exposed › error-handler', () => {
       errors: [{
         detail: 'Something bad happened',
         status: 666,
-        name: 'SomethingWrongHappenedError',
+        name: 'Error',
       }],
     }]);
   });
+
+  describe('with an error that supports data (payload)', () => {
```
ghusse, line 128: Can you add a tests with a VError ? I don't think a case with an error with an array of errors in it.

test/services/exposed/error-handler.test.js
Thenkei: 🙈

test/services/exposed/error-handler.test.js
Thenkei: Do you talk about retrieving data from VError? I've removed the code, It's not supported anymore as we only have 2 errors using VError. IMO that thing should be removed ASAP.

ghusse: My point was that the coded that is added by this PR works for errors having a property "error" with an array of errors, and I don't see a test for this case in unit tests. I assumed that it was the case of `VError`, but I'm not sure.

Is it possible to add tests for all the cases that are supported by the code, including the property `error`?

Thenkei: I forgot about that one.

`VError` cannot transport data property if that is what you meant. But I can add the support for this even if I would prefer simply removing `VError`.

ghusse (approved):

ghusse (approved):

PR #1020, opened 2023-11-02
fix(serialiser): serialiser does not compute hidden smart fields anymore

SteveBunlon:
### General

- [x] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [x] Consider the security impact of the changes made


Thenkei (approved):
Clever

SteveBunlon:
I have checked that the typings are correct on FES and FEM, which is the case already, no need to update them.

arnaudbesnier (approved):
The code looks good to me.
The PR title is a bit misleading as the prefix is "fix" and the commit describes something as a new feature.

To me it is either a feature ("developers can now specify the fields they want to serialise") or a fix ("smart collections do not compute hidden fields/columns").

PR #1045, opened 2025-01-08
fix: update forestadmin libraries

BeLi4L:
Especially to have the TTLCache fix from forestadmin-client: https://github.com/ForestAdmin/agent-nodejs/pull/1238

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
   "main": "dist/index.js",
   "dependencies": {
     "@babel/runtime": "7.19.0",
-    "@forestadmin/context": "1.31.0",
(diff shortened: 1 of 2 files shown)


BeLi4L:
The tests fail with the following error: `Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag` which doesn't look related to this MR.

Any clue?

arnaud-moncel:
Yeah Yeah, it's because you don't be part of the organization, but don't worry i'm on it.

arnaud-moncel (approved):

PR #1046, opened 2025-01-28
fix: update forestadmin-client to fix the timeout/cache issue

BeLi4L:
To have the cache fix from https://github.com/ForestAdmin/agent-nodejs/pull/1248


arnaud-moncel:
Oops din't see 😅 duplicated of #1047

PR #1059, opened 2026-06-05
fix(composite-pk): normalize fields[] query param to comma-separated string [PRD-462]

christophebrun-forest:
## Summary

On collections with a **composite primary key** requested through the **workflow executor**, `getData` (list) crashes with `TypeError: ... .split is not a function`:

- `forest-express` → `ParamsFieldsDeserializer`
- `forest-express-sequelize` → `requested-fields-extractor` (reads the same `request.query` reference)

### Root cause

The `fields[collection]` query param is documented as a **comma-separated string** (`fields[book]=id,title`). For composite-PK collections the workflow executor serializes it so that `qs` parses it into an **array** (e.g. `{ dailySpecials: ['dayOfWeek','restaurantId'] }`). Both consumers call `.split(',')` on it → crash. The browser path is unaffected (frontend sends the CSV form).

## Fix

Normalize once at the HTTP boundary so every downstream consumer — forest-express serializers **and** the ORM adapter (via the shared `request.query` reference) — receives the documented CSV shape, without patching forest-express-sequelize:

- **`src/middlewares/query-fields-normalizer.js`** (new): joins each `fields[x]` array → comma-separated string, mounted in `src/index.js` right after `bodyParser.json()`.
- **`src/deserializers/params-fields.js`**: tolerate array values (defense in depth for programmatic / exposed-service paths).
- Tests: middleware (5 cases) + deserializer (4 cases). Lint OK.

## Tickets

- Fixes [PRD-462](https://linear.app/forestadmin/issue/PRD-462)
- Related to [PRD-450](https://linear.app/forestadmin/issue/PRD-450) (composite-PK order/format for the executor). **Distinct bug**: PRD-450's forest-express twin is the composite *record-id format* (`parse_composite_id`, pipe vs JSON); this PR is the *field-selection param shape* (`fields[]` array vs CSV).

## Test plan

- [x] Unit tests green + lint
- [ ] E2E: executor `getData` on a composite-PK collection (e.g. `dailySpecials`) → no more `.split is not a function`; browser list still OK

> [!NOTE]
> ### Fix `fields[]` query param array normalization in composite PK requests
> - Adds a new [`query-fields-normalizer`](https://github.com/ForestAdmin/forest-express/pull/1059/files#diff-8152a89d93338450132d36e518d0adb382bfcd5a8f70442db5b9bf1cc89bb937) middleware that converts `fields[collection][]` array values into comma-separated strings before downstream handlers run.
> - Updates [`ParamsFieldsDeserializer.perform`](https://github.com/ForestAdmin/forest-express/pull/1059/files#diff-b9c44ef06fefeaf9699d8951b3f03336410e86459d23fd8443f4caa6ec73ccd3) to handle array inputs by flattening comma-separated entries, rather than unconditionally calling `.split(',')` which threw on non-string values.
> - The normalizer middleware is mounted in [`src/index.js`](https://github.com/ForestAdmin/forest-express/pull/1059/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556) before authentication and route handlers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ad08adf.</sup>
> 
>

### DIFF
--- a/src/deserializers/params-fields.js
+++ b/src/deserializers/params-fields.js
@@ -2,7 +2,23 @@ function ParamsFieldsDeserializer(paramsFields) {
   this.perform = () => {
     if (paramsFields) {
       return Object.keys(paramsFields).reduce((fields, modelName) => {
-        fields[modelName] = paramsFields[modelName].split(',');
+        const value = paramsFields[modelName];
+
+        if (Array.isArray(value)) {
+          // NOTICE: Some callers (e.g. collections with a composite primary key
+          //         requested through the workflow executor) serialize fields as
+          //         `fields[model][]=a&fields[model][]=b`, which qs parses into an
+          //         array instead of a comma-separated string. Entries may still
+          //         contain commas, so re-split and flatten them.
+          fields[modelName] = value.flatMap((entry) =>
+            (typeof entry === 'string' ? entry.split(',') : entry));
+        } else if (typeof value === 'string') {
+          fields[modelName] = value.split(',');
+        } else {
+          // NOTICE: Unexpected shape: keep as-is rather than throwing.
+          fields[modelName] = value;
+        }
+
         return fields;
       }, {});
     }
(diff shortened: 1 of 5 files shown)


matthv (approved):
LGTM

PR #1060, opened 2026-06-22
feat(agent): forward workflow executor routes generically [PRD-567]

matthv:
## What

4th runtime of PRD-567 (Node v1 liana, `forest-express`). Replaces the two explicit `show`/`trigger` proxy routes with a **single catch-all** that forwards every verb and sub-path under `/_internal/workflow-executions/*` to the executor `/runs` namespace. The agent no longer needs a change when the executor adds a route.

Mirrors the other three runtimes:
- Node core ([agent-nodejs#1666](https://github.com/ForestAdmin/agent-nodejs/pull/1666))
- Ruby v2 ([agent-ruby#319](https://github.com/ForestAdmin/agent-ruby/pull/319))
- Ruby v1 liana ([forest-rails#777](https://github.com/ForestAdmin/forest-rails/pull/777))

## Changes

- **Catch-all route** (`app.all('.../*')`) replacing the 2 per-route entries — all verbs.
- **Forward all request headers** except hop-by-hop / host / content-length (was an `authorization`+`cookie` whitelist).
- **Forward all response headers** except hop-by-hop — so executor-set headers (e.g. the version gate) survive the proxy.
- **Single large timeout** (no per-route values).
- **Path-traversal guard**: the wildcard can only map into `/runs` — `..`, encoded dots, backslash, null byte and leading `/` are rejected with 404, keeping non-`/runs` executor routes unreachable through the proxy.

## Tests

Rewritten suite → generic forwarding (arbitrary verb/sub-path), request + response header filtering, **path-traversal security boundary**, status/error passthrough, lazy mount. **13/13 green**.

## Known limitation (by design)

This legacy liana is **JSON-only**: responses are normalized to JSON (`res.json(...)`). A hypothetical non-JSON executor response (e.g. `text/plain`) would not pass through verbatim. The executor only returns JSON today, so this does not affect the PRD-567 contract in practice; the modern `@forestadmin/agent` path (agent-nodejs#1666) handles raw passthrough.

fixes PRD-567

> [!NOTE]
> ### Replace fixed workflow executor routes with a catch-all proxy for all verbs and sub-paths
> - Replaces two explicit routes (`GET /:runId` and `POST /:runId/trigger`) with a single `app.all('/_internal/workflow-executions/*', ...)` that proxies any HTTP method and sub-path to the workflow executor.
> - Broadens header forwarding: outgoing requests now include all client headers except hop-by-hop, `Host`, and encoding/framing headers; executor response headers are relayed back to the client under the same exclusion rules.
> - Adds path traversal protection via `executorPath()`, which rejects paths containing `..`, `%2e`, `\`, or null bytes with a 404.
> - Increases proxy timeouts from 30s/60s to 120s response/deadline in [workflow-executor.js](https://github.com/ForestAdmin/forest-express/pull/1060/files#diff-6b172fbe49f428449c9d9e08b79754ed108880ed5d174d9bfc9352a179a987ef).
>
> #### 🖇️ Linked Issues
> Implements [PRD-567](ticket:linear/PRD-567).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b96c512.</sup>
> 
>

### DIFF
--- a/src/routes/workflow-executor.js
+++ b/src/routes/workflow-executor.js
@@ -3,43 +3,72 @@ const superagent = require('superagent');
 const path = require('../services/path');
 const auth = require('../services/auth');
 
-const FORWARDED_REQUEST_HEADERS = ['authorization', 'cookie'];
+const EXECUTOR_PREFIX = '/runs';
+// Hop-by-hop / client-recomputed headers — never forwarded (request or response side).
+const SKIPPED_HEADERS = new Set([
+  'connection', 'keep-alive', 'transfer-encoding', 'upgrade', 'te', 'trailer',
+  'proxy-authenticate', 'proxy-authorization', 'host', 'content-length',
+]);
+const UNSAFE_PATH_FRAGMENTS = ['..', '%2e', '%2E', '\\', '\0'];
 
 // NOTICE: Bound the proxied request so a hanging executor cannot tie up the
 //         connection indefinitely. A timeout raises an error without `.response`,
 //         so it falls through to the 503 branch.
-const REQUEST_TIMEOUT = { response: 30000, deadline: 60000 };
+const REQUEST_TIMEOUT = { response: 120000, deadline: 120000 };
 
-function pickForwardedHeaders(requestHeaders) {
+function forwardedRequestHeaders(requestHeaders) {
   const headers = {};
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
LGTM

PR #1061, opened 2026-06-25
feat(agent): forward any executor route generically [PRD-567]

matthv:
## What

Ports the generic executor passthrough to forest-express (legacy Node liana), matching the agent-nodejs reference ([agent-nodejs#1707](https://github.com/ForestAdmin/agent-nodejs/pull/1707), released 1.81.0).

- **Single catch-all mount** `_internal/executor/*` → executor `/*` **verbatim** (all verbs). Replaces the `/runs`-confined proxy. Callers address runs as `runs/<id>`, so any executor route (mcp-oauth-credentials, future ones) is reachable with no agent change.

## Security

- **SSRF guard**: reject empty / leading-`/` / `..` / `%2e`/`%2E` / backslash / NUL, and verify the resolved URL never leaves the executor origin.
- **`.redirects(0)`**: don't follow executor redirects — a 3xx `Location` to an off-origin host can't make the agent reach another host (matches the Node agent's raw http client, which never follows redirects).

## Tests

**15/15**. Run GET (caller includes `runs/`), POST trigger, **non-runs route verbatim** (asserts no `/runs` prefix injected), 4xx passthrough, 503 unreachable, req/resp header filtering (gzip + accept-encoding), SSRF traversal vectors → 404, **redirect not followed**, lazy mount (`/forest/_internal/executor/*`).

## Note
Same design as Node, one safe nuance: a leading control-char (`\t//host`) is *forwarded* here (string-concat keeps the host fixed → executor 404s it) rather than rejected as in Node (`new URL` relative resolution) — both stay on the executor origin.

> [!NOTE]
> ### Forward arbitrary executor routes generically under `/forest/_internal/executor/*`
> - Changes the catch-all route from `/forest/_internal/workflow-executions/*` to `/forest/_internal/executor/*` in [workflow-executor.js](https://github.com/ForestAdmin/forest-express/pull/1061/files#diff-6b172fbe49f428449c9d9e08b79754ed108880ed5d174d9bfc9352a179a987ef).
> - Removes the automatic `/runs` prefix injection so sub-paths are forwarded verbatim; callers targeting run resources must now include `runs/` themselves.
> - Adds an SSRF guard that returns 404 if the constructed upstream URL's origin differs from the executor's base origin.
> - Behavioral Change: the proxy no longer follows upstream 3xx redirects — redirect responses are returned directly to the client.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized fce90cd. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/src/routes/workflow-executor.js
+++ b/src/routes/workflow-executor.js
@@ -3,7 +3,6 @@ const superagent = require('superagent');
 const path = require('../services/path');
 const auth = require('../services/auth');
 
-const EXECUTOR_PREFIX = '/runs';
 // Never forwarded (request or response): hop-by-hop, Host, and body-framing headers.
 // res.json() re-serializes the body, so upstream length/encoding no longer match — and
 // forwarding accept-encoding would relay an encoding the re-emitted body doesn't use.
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
Goot To Me

